PR #9, opened 2018-11-14
Extract vet and lint into separate make targets

aduffeck:


bin/lint
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set-i
```
manno: Is this vaild?

PR #11, opened 2018-11-14
Switch to cobra,zap

manno:
Add cobra/viper for command line processing and add more tests. Inject zap logger.


e2e/cli_test.go
```diff
@@ -0,0 +1,81 @@
...
+		cmd := exec.Command(cliPath, arg...)
+		session, err = gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
+		return
+	}
+
+	Describe("help", func() {
+		It("should show the help for server", func() {
+			session, err := act("help")
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(session.Out).Should(gbytes.Say(`Usage:`))
```
aduffeck: How about dot-importing `gbytes` and getting rid of this qualifier just like we already do with `gomega` and its matchers? Would make the expectations read more naturally imo.

e2e/cli_test.go
```diff
@@ -0,0 +1,81 @@
...
+			Eventually(session.Out).Should(gbytes.Say(`-h, --help`))
+			Eventually(session.Out).Should(gbytes.Say(`-c, --kubeconfig string`))
```
aduffeck: Is there any value in testing the output for all of the currently available options and commands?
I think it's good to test that the command works in general, but I can't really think of a case where testing all of its output would catch any real issue.

integration/integration_suite_test.go
```diff
@@ -60,21 +60,22 @@ var _ = AfterSuite(func() {
...
-	s.VersionedClientset, err = versioned.NewForConfig(kubeConfig)
-	return
-}
-
-func (s *IntegrationSuite) getTestNamespace() string {
+func (s *IntegrationSuite) setup() {
 	ns, found := os.LookupEnv("TEST_NAMESPACE")
 	if !found {
-		return "default"
+		s.namespace = "default"
```
aduffeck: Shouldn't this say

```suggestion
		ns = "default"
```

e2e/cli_test.go
manno: I want to make sure that all options are there and no additional options have been added without looking at the tests. I'll adapt this test, so it's more like the 'shows all available commands'.

PR #17, opened 2018-11-29
Refactor test cases for the resolver_test.go

qu1queee:
Reason behind:
- Running unit tests using the `fake_interpolator.go` functions, did not resolved 
   a proper final manifest, when using a more complex `bosh-manifest` in combination 
   with `ops-files`.
- Deleted the `fake_interpolator.go`, and move to use directly the
  `interpolator.go` functions.
- Also, add unit test to cover both types for ops-files, `remove` and `replace`.

I´m not sure if removing the `fake_interpolator.go` is desired.


qu1queee:
I just realized this could be invalid, due to the test cases already in `interpolator_test.go`, pls close it if this is the case.

edwardstudy:
Yes, @qu1queee. We thought each package test only covers self functionality and their dependencies already have desired test cases. As you said, if we want to more test cases, we should add they in `interpolator_test.go`

edwardstudy:
Closed

PR #25, opened 2018-12-04
Add interface for a credentials generator

aduffeck:
[161745020](https://www.pivotaltracker.com/story/show/161745020)

### DIFF
--- a/pkg/credsgen/generator.go
+++ b/pkg/credsgen/generator.go
@@ -1,0 +1,42 @@
+package credsgen
+
+// PasswordGenerationRequest specifies the generation parameters for Passwords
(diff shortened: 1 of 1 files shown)


aduffeck:
@viovanov Hm, I think `keygen` doesn't fit that well because it's also about passwords and certificates. I didn't choose `secrets` because it's likely to be confused with k8s secrets imo. `secretgen` is slightly better, but I still feel  that `credentials` reflects best what the package is about. `credentialsgen`?

viovanov:
@aduffeck ok, let's go with `credsgen`. I agree with the concern for "secrets".

aduffeck:
@viovanov :+1:

PR #29, opened 2018-12-06
Refactor dir structure and Dockerfile

qu1queee:
1. Refactor the location of the `main.go` and cobra classes
   This include:
   - move `main.go` inside `cmd/cf-operator`, in order to follow
     generic project layout standards.
   - place the cobra classes `root.go` and `version.go` under 
    `cmd/internal`. So, they will not be next to `main.go`.
   - Refactor any reference towards the old dir locations.

2. The `cfcontainerization/cf-operator` Dockerfile, can be build by
   using the multistage, where in the first stage we build the
   cf-operator binary, and then in the final stage we insert them.
   This should improve:
   - No need to run make/build before make/image.
   - In the future, if more golang repos are needed inside the image,
     it should improve the building image time.


manno:
Adding to the previous discussion at https://cloudfoundry.slack.com/archives/GD4SSJF33/p1544100863006600

I don't see a need for moving the main.go, i.e. we don't plan on having multiple binaries. Can you link to generic project layout standard?

About the multistage docker build:
I'd prefer to keep the Dockerfile in a sub directory, but can't really give a reason why :)
Also, we vendor all dependencies with `dep ensure`, so building inside docker has the advantage that missing deps will break the build.  Not sure how this will speed up the build, though.

qu1queee:
Hi @manno ,

I think is a good practice to have the `main.go` under `cmd`, which is not related to having multiple binaries. The layout is explain in https://github.com/golang-standards/project-layout .
Regarding the Dockerfile, having a multistage approach is giving you:
- a full binary compilation inside the image, without the need of doing it outside via make.
- Not having it in a sub directory, will allow us to enable Dockerhub to build this automatically. I will need @viovanov to enable this for the cfcontainerization team. Having it in a subfolder, make it really complex.
- As you said, building the binary inside the Dockerfile, ensure yours vendor is in a good state.
- Speed was primarily focused on the fact that network related things during Docker image creation could be faster in Docker Hub, but this might be not that relevant for our Dockerfile here.

manno:
Definitely makes sense to move the Dockerfile if it makes things easier with dockerhub. 
@aduffeck does this affect our Concourse pipelines?

https://github.com/golang-standards doesn't convince me, but I'm always wary of things that look official, but are not. (I should put my vim config at github.com/default-vim/config for fun).
I see no harm in moving main.go.

Anyone else got opinions?

aduffeck:
Yes, we'll have to adapt the path in the pipeline but I can do this once this is merged, no problem.

Building automatically on dockerhub sounds nice but the problem is that we couldn't apply our versioning there so I think we'll need the pipeline to push properly versioned images anyway.

HeavyWombat:
The content of https://github.com/golang-standards  is nothing official, the Go community does not enforce a standard here. After looking at different Go projects and seeing different styles, I came to the conclusion that some kind of "standard" look and feel is nice and helps to have a clean understandable structure. For me, the one explained by the aforementioned link gives the best overview/explanation.

qu1queee:
@aduffeck for the image versioning, in `dockerhub`, there is the `Docker Tag Name` option, where you can specify the `name` for the tag, e.g. `latest`, otherwise it will take the name of the branch.

PR #34, opened 2018-12-10
Switch to go modules for dependency management

mudler:
Not sure if this is wanted at some point, but seeing also that upstream [libbuildpack](https://github.com/cloudfoundry/libbuildpack/commit/82a5d6eeddf32507716620faec9217167701952c) is moving towards Go modules, I thought that this might be of interest, since this should become the official way sooner or later.

This PR will make an hard requirement of Go >=1.11 during build (Maybe we should bail-out also when <go-1.11 is used ? ).

In Go 1.11 modules were introduced, quoting ``go help modules``:

> Go 1.11 includes preliminary support for Go modules,
> including a new module-aware 'go get' command.
> We intend to keep revising this support, while preserving compatibility,
> until it can be declared official (no longer preliminary),
> and then at a later point we may remove support for work
> in GOPATH and the old 'go get' command.

With this change, when building from a ``GOPATH`` it is required to enable ``GO111MODULE`` environment variable ( which becomes redundant if ```GO111MODULE=on``` becomes default in future Go versions )

```GO111MODULE=on go build```

Otherwise no change is required ( ``go build`` should now just work outside of ``GOPATH`` ). 
The build process now fetches automatically the required dependencies - in case of an offline scneario it's possible to generate the vendor directory by ``go mod vendor`` and reusing it later by forcing ```-mod=vendor``` while running go commands ( e.g. [ ```go build -mod=vendor```](https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away) ) .

It should not affect pipelines as long as they are not building from a GOPATH directory.

~~WIP: ```/bin/gen-kube``` and ```/bin/gen-fakes``` needs to be adapted still~~

Note:
- `` bin/gen-fakes`` does not work if called inside GOPATH


manno:
Getting rid of `vendor` in the repo would be great. I just bumped `controller-runtime` and it's a lot of files and dependencies :+1:

PR #40, opened 2018-12-11
Do not hardcode values in InMemoryGenerator

mudler:
Allows to setup different values for the generator, and adapts tests to
use the default constructor for more consistency.

Imho it's easier to create custom objects that are overriding defaults on-need.


mudler:
Sure, will do :+1:

mudler:
Just added tests for defaults now, not for custom options, so not ready yet for merge

aduffeck:
This is superseded by #52 which adds the missing tests.

PR #43, opened 2018-12-11
fixes for gomod

manno:


mudler:
Cool! Was not sure how to handle best the code-generator, I guess this can be removed as well then https://github.com/cloudfoundry-incubator/cf-operator/blob/master/.gitignore#L11

manno:
@mudler yes, I was a little surprised about the version coupling between them, but makes sense.

PR #47, opened 2018-12-12
Add event recorder for reconciler

zhangtbj:
1, Add Event recorder for Reconciler to record fail at BOSHDeployment CR level
2, Add an integration test for the event message test
3, All test passed:
```
Ran 9 of 9 Specs in 3.179 seconds
SUCCESS! -- 9 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

### DIFF
--- a/integration/deploy_test.go
+++ b/integration/deploy_test.go
@@ -55,11 +55,6 @@ var _ = Describe("Deploy", func() {
 	})
 
 	Context("when incorrectly setup", func() {
-		AfterEach(func() {
-			err := env.WaitForPodsDelete(env.Namespace)
-			Expect(err).NotTo(HaveOccurred())
-		})
-
 		It("should report correct error if ops for is incorrect", func() {
 			tearDown, err := env.CreateConfigMap(env.Namespace, env.DefaultBOSHManifest("manifest"))
 			Expect(err).NotTo(HaveOccurred())
@@ -73,13 +68,12 @@ var _ = Describe("Deploy", func() {
 			Expect(err).NotTo(HaveOccurred())
 			defer tearDown()
 
-			_, tearDown, err = env.CreateBOSHDeployment(env.Namespace, env.InterpolateBOSHDeployment("test", "manifest", "bosh-ops", "bosh-ops-secret"))
+			boshDeployment, tearDown, err := env.CreateBOSHDeployment(env.Namespace, env.InterpolateBOSHDeployment("test", "manifest", "bosh-ops", "bosh-ops-secret"))
 			Expect(err).NotTo(HaveOccurred())
 			defer tearDown()
 
-			// check for pod
-			pass, result := env.HasCREvent(env.Namespace)
-			Expect(pass).To(BeTrue())
+			success, result := env.GetBOSHDeploymentEventMessage(env.Namespace, boshDeployment.ObjectMeta.Name, string(boshDeployment.ObjectMeta.UID))
+			Expect(success).To(BeTrue())
 			Expect(result).To(ContainSubstring("Expected to find exactly one matching array item for path '/instance-groups/name=api' but found 0"))
 		})
 
--- a/integration/deploy_test.go
+++ b/integration/deploy_test.go
@@ -55,7 +55,7 @@ var _ = Describe("Deploy", func() {
 	})
 
 	Context("when incorrectly setup", func() {
-		It("should report correct error if ops for is incorrect", func() {
+		It("failed to deploy if ResolveCRD error occurred", func() {
 			tearDown, err := env.CreateConfigMap(env.Namespace, env.DefaultBOSHManifest("manifest"))
 			Expect(err).NotTo(HaveOccurred())
 			defer tearDown()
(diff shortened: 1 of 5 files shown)


zhangtbj:
We will also provide a doc about how to use event recorder and a common solution for all components by using another story:
https://www.pivotaltracker.com/n/projects/2192232

integration/deploy_test.go
```diff
@@ -55,6 +55,28 @@ var _ = Describe("Deploy", func() {
 	})
 
 	Context("when incorrectly setup", func() {
+		It("should report correct error if ops for is incorrect", func() {
```
manno: I don't understand the description, "ops for is incorrect". Ops for what :) ?

integration/deploy_test.go
```diff
@@ -55,6 +55,28 @@ var _ = Describe("Deploy", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			defer tearDown()
+
+			boshDeployment, tearDown, err := env.CreateBOSHDeployment(env.Namespace, env.InterpolateBOSHDeployment("test", "manifest", "bosh-ops", "bosh-ops-secret"))
+			Expect(err).NotTo(HaveOccurred())
+			defer tearDown()
+
+			success, result := env.GetBOSHDeploymentEventMessage(env.Namespace, boshDeployment.ObjectMeta.Name, string(boshDeployment.ObjectMeta.UID))
+			Expect(success).To(BeTrue())
+			Expect(result).To(ContainSubstring("Expected to find exactly one matching array item for path '/instance-groups/name=api' but found 0"))
```
manno: Does this test one of the events from `reconciler.go`? Which one?

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -66,25 +69,30 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
+		r.recorder.Event(instance, corev1.EventTypeWarning, "GetCRD Error", err.Error())
```
manno: This one is missing a test?

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -66,25 +69,30 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
 		return reconcile.Result{}, err
 	}
 
 	// retrieve manifest
 	manifest, err := r.resolver.ResolveCRD(instance.Spec, request.Namespace)
 	if err != nil {
+		r.recorder.Event(instance, corev1.EventTypeWarning, "ResolveCRD Error", err.Error())
```
manno: I think this is tested

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -66,25 +69,30 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
 		return reconcile.Result{}, err
 	}
 
 	// TODO validation
 	if len(manifest.InstanceGroups) < 1 {
-		return reconcile.Result{}, fmt.Errorf("manifest is missing instance groups")
+		err = fmt.Errorf("manifest is missing instance groups")
+		r.recorder.Event(instance, corev1.EventTypeWarning, "MissingInstance Error", err.Error())
```
manno: This should have a test. We can also remove the TODO comment.

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -66,25 +69,30 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+		return reconcile.Result{}, err
 	}
 
 	// Define a new Pod object
 	pod := newPodForCR(manifest, request.Namespace)
 
 	// Set BOSHDeployment instance as the owner and controller
 	if err := r.setReference(instance, pod, r.scheme); err != nil {
+		r.recorder.Event(instance, corev1.EventTypeWarning, "NewPodForCR Error", err.Error())
```
manno: I don't think any of the 'new pod' events needs a test. The pod functionality is a left over and will probably be removed in the future. 
The BOSHDeployment controller will serve as an entry point, to create the other resources (extended*). But I don't know which story that belongs to and I'm hesitant to remove the newPodCR code and existing tests.

integration/deploy_test.go
edwardstudy: It tested this event: https://github.com/cloudfoundry-incubator/cf-operator/blob/f5abafcd7742f3eeecef3b4305279f1140ce18b6/pkg/kube/controllers/boshdeployment/reconciler.go#L79

I'll update other events

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: I think so.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: Updated test cases

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: Updated test cases

PR #57, opened 2018-12-17
List all StatefulSet versions for an ExtendedStatefulSet and detecting if it's running or not

edwardstudy:
Hi, 

We had implemented basic versions listing for first two tasks: https://www.pivotaltracker.com/story/show/162594634 

Please help to review, and we are updating integration tests in the meantime.

Best,

@edwardstudy && @zhangtbj, IBM


pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -258,3 +282,77 @@ func (r *ReconcileExtendedStatefulSet) getActualStatefulSet(ctx context.Context,
...
+	)
+	if err != nil {
+		return false, err
+	}
+
+	for _, pod := range podList.Items {
+		if metav1.IsControlledBy(&pod, statefulSet) {
+			if isPodReady(&pod) {
+				r.log.Debug("Pod '", statefulSet.Name, "' owned by StatefulSet '", statefulSet.Name, "' is running.")
+				return true, nil
```
aduffeck: Shouldn't this only return `true` if **all** pods of the StatefulSet are ready?

pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -258,3 +282,77 @@ func (r *ReconcileExtendedStatefulSet) getActualStatefulSet(ctx context.Context,
...
+		}
+	}
+
+	return false, nil
+}
+
+// isPodReady returns false if the Pod Status is nil
+func isPodReady(pod *v1.Pod) bool {
+	var condition *v1.PodCondition
+	for i := range pod.Status.Conditions {
```
aduffeck: Could this maybe be simplified using [GetPodCondition](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/pod/util.go#L194)?

pkg/kube/controllers/extendedstatefulset/reconciler_test.go
```diff
@@ -0,0 +1,83 @@
...
+		core, _ := observer.New(zapcore.InfoLevel)
+		log = zap.New(core).Sugar()
+	})
+
+	JustBeforeEach(func() {
+		reconciler = cfd.NewReconciler(log, manager, controllerutil.SetControllerReference)
+	})
+
+	Describe("Reconcile", func() {
+		Context("when the manifest", func() {
```
aduffeck: This context doesn't make sense, I'd remove it.

pkg/kube/controllers/extendedstatefulset/reconciler_test.go
```diff
@@ -0,0 +1,83 @@
...
+			var (
+				client *cfakes.FakeClient
+			)
+			BeforeEach(func() {
+				client = &cfakes.FakeClient{}
+				client.GetStub = func(i context.Context, name types.NamespacedName, object runtime.Object) error {
+					object = &exsts.ExtendedStatefulSet{
```
aduffeck: I don't think this assignment actually changes anything. The test should pass just fine without it. But then I wonder why it passes if there is no ExtendedStatefulSet...

pkg/kube/controllers/extendedstatefulset/reconciler_test.go
```diff
@@ -0,0 +1,83 @@
...
+					}
+
+					return nil
+				}
+
+				manager.GetClientReturns(client)
+				manager.GetSchemeReturns(scheme.Scheme)
+			})
+
+			It("create new stateful set and requeue for extended stateful set", func() {
```
aduffeck: The test doesn't actually verify that a StatefulSet was created (or rather that the fake client was instructed to create one), I think it would be good to add that.

pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -89,10 +92,13 @@ func (r *ReconcileExtendedStatefulSet) Reconcile(request reconcile.Request) (rec
 		// If it doesn't exist, create it
 		r.log.Info("StatefulSet '", desiredStatefulSet.Name, "' for ExtendedStatefulSet '", request.NamespacedName, "' not found, will be created.")
 
+		// Reserve user-defined template
+		reservedTemplate := exStatefulSet.Spec.Template.DeepCopy()
```
aduffeck: Why is this needed?

pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -258,3 +282,77 @@ func (r *ReconcileExtendedStatefulSet) getActualStatefulSet(ctx context.Context,
...
+			return result, errors.Wrapf(err, "Version annotation is not an int: %s", strVersion)
+		}
+
+		ready, err := r.isStatefulSetReady(ctx, &statefulSet)
+		if err != nil {
+			return nil, err
+		}
+
+		if ready {
+			result[version] = true
```
aduffeck: It may be a matter of taste but I think I would assign `ready` directly.

pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -258,3 +282,77 @@ func (r *ReconcileExtendedStatefulSet) getActualStatefulSet(ctx context.Context,
 
 	return result, maxVersion, nil
 }
+
+// listStatefulSetVersions gets all StatefulSets' versions and ready status owned by the ExtendedStatefulSet
+func (r *ReconcileExtendedStatefulSet) listStatefulSetVersions(ctx context.Context, exStatefulSet *essv1a1.ExtendedStatefulSet) (map[int]bool, error) {
```
aduffeck: I know that `listStatefulSets` was already defined on `*ReconcileExtendedStatefulSet`, but I wonder if it would make sense to move these functions to the `ExtendedStatefulSet` type instead. They are not really `Reconciler` domain imo.

pkg/kube/controllers/extendedstatefulset/reconciler.go
edwardstudy: From story task description, **at least one** pod is running should indicate the statefulSet is available. CMIIW

pkg/kube/controllers/extendedstatefulset/reconciler.go
edwardstudy: Hi, @aduffeck. I found that exStatefulSet will obtain three default values: `ImagePullPolicy `, `TerminationMessagePath ` and `TerminationMessagePolicy ` if `spec.template` doesn't speficy after creating its statefulSet: https://github.com/kubernetes/kubernetes/blob/77a1199b6cba0df666d0a830483e4c725fe3b444/pkg/apis/core/v1/defaults.go#L77-L95

Then controller will update exStatefulSet with inconsistent template sha1sum. When reconcile loop continue, reconcile would create next version because `spec.template` has been changed.

So I just keep `spec.template`  which I mean is user-defined.

edwardstudy:
Weird: 

```
:cf-operator bjxzi$ ginkgo -p -r --randomizeAllSpecs -failOnPending --trace -skipPackage integration,e2e
Will skip:
  ./e2e
  ./integration
[1545301266] Manifest Suite - 40 specs - 7 nodes •••••••••••••••••••••••••••••••••••••••• SUCCESS! 36.599277ms
[1545301266] InMemoryGenerator Suite - 13 specs - 7 nodes •••••••••
------------------------------
• [SLOW TEST:6.308 seconds]
InMemoryGenerator GenerateCertificate when generating a certificate fails if the passed CA is not a CA
/Volumes/journey/workspace/src/code.cloudfoundry.org/cf-operator/pkg/credsgen/in_memory_generator/certificate_test.go:46
------------------------------
• [SLOW TEST:7.654 seconds]
InMemoryGenerator GenerateCertificate when generating a certificate considers the alternative names
/Volumes/journey/workspace/src/code.cloudfoundry.org/cf-operator/pkg/credsgen/in_memory_generator/certificate_test.go:70
------------------------------
• [SLOW TEST:8.441 seconds]
InMemoryGenerator GenerateRSAKey generates an RSA key
/Volumes/journey/workspace/src/code.cloudfoundry.org/cf-operator/pkg/credsgen/in_memory_generator/rsa_key_test.go:22
------------------------------
• [SLOW TEST:8.738 seconds]
InMemoryGenerator GenerateCertificate when generating a certificate considers the common name
/Volumes/journey/workspace/src/code.cloudfoundry.org/cf-operator/pkg/credsgen/in_memory_generator/certificate_test.go:58
------------------------------
 SUCCESS! 8.767910968s
[1545301266] Controllers Suite - 1 specs - 7 nodes • SUCCESS! 14.32892ms
[1545301266] Boshdeployment Suite - 6 specs - 7 nodes •••••• SUCCESS! 23.121216ms
[1545301266] ExtendedStatefulSet Suite - 4 specs - 7 nodes •••• SUCCESS! 24.578699ms

Ginkgo ran 5 suites in 16.878426429s
Test Suite Passed
```

pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -258,3 +283,61 @@ func (r *ReconcileExtendedStatefulSet) getActualStatefulSet(ctx context.Context,
...
+
+	statefulSets, err := r.listStatefulSets(ctx, exStatefulSet)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, statefulSet := range statefulSets {
+		strVersion, ok := statefulSet.Annotations[essv1a1.AnnotationVersion]
+		if !ok {
+			strVersion = "0"
```
viovanov: I am not sure this is a correct assumption.
Can you describe a case where this would happen?
I'm thinking this should be an error :/

pkg/kube/controllers/extendedstatefulset/reconciler.go
```diff
@@ -258,3 +283,61 @@ func (r *ReconcileExtendedStatefulSet) getActualStatefulSet(ctx context.Context,
...
+		result[version] = ready
+	}
+
+	return result, nil
+}
+
+// isStatefulSetReady returns true if one owned Pod is running
+func (r *ReconcileExtendedStatefulSet) isStatefulSetReady(ctx context.Context, statefulSet *v1beta1.StatefulSet) (bool, error) {
+	podList := &api.PodList{}
+	err := r.client.List(
```
viovanov: Is there a way to get only the pods we own?
Seems inefficient to get ALL pods in the namespace.

pkg/kube/controllers/extendedstatefulset/reconciler.go
viovanov: Yeah, if one pod is running, we're good.

viovanov:
@aduffeck @edwardstudy any ideas why tests are failing?

edwardstudy:
@viovanov It was affected by my code that I updated go modules. I'll check it locally again.

aduffeck:
@viovanov @edwardstudy The tests pass locally for me as well. I'm looking into it.

aduffeck:
I couldn't really figure out the exact issue but it was related to the go module cache we keep around in concourse. It somehow made go use the latest version of all dependencies instead of the ones specified in `go.mod`. Purging the cache made this problem go away...

edwardstudy:
Updated commits. Thanks for review.

pkg/kube/controllers/extendedstatefulset/reconciler_test.go
edwardstudy: Updated

pkg/kube/controllers/extendedstatefulset/reconciler.go
edwardstudy: Use label selector to filer owned pods. I thought we can modify CRD and apply this selector to ex StatefulSet with its owned statefulSet.

pkg/kube/controllers/extendedstatefulset/reconciler.go
edwardstudy: Return error directly. I thought it again and statefulSet should have this annotation otherwise it had some inconsistent issue.

edwardstudy:
Updated docs and integration tests.

edwardstudy:
Hi, @viovanov. I had already updated code. Would you please continue to review if you have time?

PR #59, opened 2018-12-17
Introduce ManifestPersister interface (tracker ID 161744952)

HeavyWombat:
Introduce `ManifestPersister` interface to enable exploiters to perform the following actions:
- creating a desired manifest secret
- deleting a desired manifest secret
- creating a new version of a desired manifest
- decorating the secret with key/value labels
- listing all versions
- retrieving a version of a desired manifest
- retrieving the latest version of a desired manifest

https://www.pivotaltracker.com/story/show/161744952

In order to keep the interface consistent and compact, the function names only refer to what is actually persisted: the manifest. The implementation detail that is is stored as a `secret` is visible as the return type as well as in the godoc texts.

### DIFF
--- a/pkg/kube/secrets/manifest.go
+++ b/pkg/kube/secrets/manifest.go
@@ -5,12 +5,11 @@ import (
 	"regexp"
 	"strconv"
 
+	"k8s.io/client-go/kubernetes"
+
 	"code.cloudfoundry.org/cf-operator/pkg/bosh/manifest"
 	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v2"
-
-	"k8s.io/client-go/kubernetes"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -32,42 +31,40 @@ const (
 //
 // The version number is an integer that is incremented with each version of
 // the manifest, which the greatest number being the current/latest version.
+//
+// When persisting a new manifest, a source description is required, which
+// should explain the sources of the rendered manifest, e.g. the location of
+// the Custom Resource Definition that generated it.
 type ManifestPersister interface {
-	// Creates a new secret version of the provided manifest and a description of the "sources" used to render the manifest (e.g. the location of the CRD that generated it)
 	PersistManifest(manifest manifest.Manifest, sourceDescription string) (*corev1.Secret, error)
-
-	// Deletes all persisted versions of the manifest
 	DeleteManifest() error
-
-	// Adds a label to the current/latest version on the manifest
 	DecorateManifest(key string, value string) (*corev1.Secret, error)
-
-	// Lists all persisted versions of the manifest
 	ListAllVersions() ([]corev1.Secret, error)
-
-	// Get a specific version of the manifest
 	RetrieveVersion(version int) (*corev1.Secret, error)
-
-	// Get the current/latest version of the manifest
 	RetrieveLatestVersion() (*corev1.Secret, error)
 }
 
-type manifestPersisterImpl struct {
+// ManifestPersisterImpl contains the required fields to persist a manifest
+type ManifestPersisterImpl struct {
 	client         kubernetes.Interface
 	namespace      string
 	deploymentName string
 }
 
-// NewManifestPersister returns a ManifestPersister implementation to be used when working with desired manifest secrets
-func NewManifestPersister(client kubernetes.Interface, namespace string, deploymentName string) ManifestPersister {
-	return &manifestPersisterImpl{
+// NewManifestPersister returns a ManifestPersister implementation to be used
+// when working with desired manifest secrets
+func NewManifestPersister(client kubernetes.Interface, namespace string, deploymentName string) ManifestPersisterImpl {
+	return ManifestPersisterImpl{
 		client,
 		namespace,
 		deploymentName,
 	}
 }
 
-func (p *manifestPersisterImpl) PersistManifest(manifest manifest.Manifest, sourceDescription string) (*corev1.Secret, error) {
+// PersistManifest creates a new version of the manifest if it already exists,
+// or a first one it not. A source description should explain the sources of
+// the rendered manifest, e.g. the location of the CRD that generated it
+func (p ManifestPersisterImpl) PersistManifest(manifest manifest.Manifest, sourceDescription string) (*corev1.Secret, error) {
 	currentVersion, err := getGreatestVersion(p)
 	if err != nil {
 		return nil, err
@@ -101,7 +98,8 @@ func (p *manifestPersisterImpl) PersistManifest(manifest manifest.Manifest, sour
 	})
 }
 
-func (p *manifestPersisterImpl) RetrieveVersion(version int) (*corev1.Secret, error) {
+// RetrieveVersion returns a specific version of the manifest
+func (p ManifestPersisterImpl) RetrieveVersion(version int) (*corev1.Secret, error) {
 	list, err := p.ListAllVersions()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/kube/secrets/manifest.go
```diff
@@ -0,0 +1,241 @@
...
+	// Creates a new secret version of the provided manifest and a description of the "sources" used to render the manifest (e.g. the location of the CRD that generated it)
+	PersistManifest(manifest manifest.Manifest, sourceDescription string) (*corev1.Secret, error)
+
+	// Deletes all persisted versions of the manifest
+	DeleteManifest() error
+
+	// Adds a label to the current/latest version on the manifest
+	DecorateManifest(key string, value string) (*corev1.Secret, error)
+
+	// Lists all persisted versions of the manifest
```
manno: I feel like some of these comments are redundant? But it's fine if you want to keep them.

pkg/kube/secrets/manifest.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+type manifestPersisterImpl struct {
+	client         kubernetes.Interface
+	namespace      string
+	deploymentName string
+}
+
+// NewManifestPersister returns a ManifestPersister implementation to be used when working with desired manifest secrets
+func NewManifestPersister(client kubernetes.Interface, namespace string, deploymentName string) ManifestPersister {
```
manno: * returns interface
I would have returned the implementation instead of the interface, it still fulfills the contract. Of course that only works if your implementation is public. Your approach has the advantage of providing very tight, Java-ish, encapsulation :)
I remember running into problems testing structs like that in the past, but that might have been me. Anyhow, I enjoyed reading https://mycodesmells.com/post/accept-interfaces-return-struct-in-go
* `uses kubernetes.Interface`
The controllers use a wrapped client from `controller-runtime`, it's not too different. I mostly use it, because I'm under the impression it should provide caching. Not sure where this feature will be used and what client it should use.
* returning a pointer
As far as I understand it's beneficial to return a pointer if the struct uses a pointer receiver. However I'm not sure the `ManifestPersisterImpl` should use a pointer receiver :)  It doesn't look like any of the fields are being mutated, so it could get by with a regular receiver?

pkg/kube/secrets/manifest.go
```diff
@@ -0,0 +1,241 @@
...
+		labels = make(map[string]string)
+	}
+
+	labels[key] = value
+	secret.SetLabels(labels)
+
+	return p.client.CoreV1().Secrets(p.namespace).Update(secret)
+}
+
+func getGreatestVersion(p *manifestPersisterImpl) (int, error) {
```
manno: Why not define the private functions on the struct?

pkg/kube/secrets/manifest.go
```diff
@@ -0,0 +1,241 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	secret, err := p.client.CoreV1().Secrets(p.namespace).Get(secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	labels := secret.GetLabels()
```
manno: I read up on labels vs annotations https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ and it seems we should be using labels for selecting and user facing stuff. An automatic version number used by our service sound like a case for annotations?

pkg/kube/secrets/manifest_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		return manifest
+	}
+
+	exampleSourceDescription := "created by a unit-test"
+
+	var namespace = "default"
+	var deploymentName = "fake-deployment"
+
+	Context("Creating a desired manifest secret", func() {
```
manno: Interesting. I would have structured the ginkgo unit test it like 

```
Describe 'PersistManifest'
  Context 'when there is no versioned secret'
     BeforeEach (to setup test data)
     It 'creates a secret' (to make the assertion)
Describe 'ListAllVersion'
...
```

pkg/kube/secrets/manifest_test.go
```diff
@@ -0,0 +1,198 @@
...
+var _ = Describe(`Versioning and persistence for "desired manifests"`, func() {
+	exampleManifest := func(data string) manifest.Manifest {
+		var manifest manifest.Manifest
+		err := yaml.Unmarshal([]byte(data), &manifest)
+		Expect(err).ToNot(HaveOccurred())
+
+		return manifest
+	}
+
+	exampleSourceDescription := "created by a unit-test"
```
manno: This doesn't match the surrounding style, should also use `var`.

pkg/kube/secrets/manifest_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	var namespace = "default"
```
manno: Multiple `var` blocks could be grouped in a block

HeavyWombat: Ok, will replace it with one `var` block.

pkg/kube/secrets/manifest_test.go
HeavyWombat: Ok, will replace it with one `var` block.

pkg/kube/secrets/manifest.go
HeavyWombat: Thanks for bringing that up. It is actually influenced by a YouTube video I saw on common mistakes in Go. Coming from long time exposure of (I am not ashamed to say it) Java, [this particular part of the video](https://youtu.be/29LLRKIL_TI?t=638) struck home. So I would be interested to hear your preference. Since it does not change any state, it felt like a helper function and therefore was written in that way. I am happy to change it to an unexported method if you want.

pkg/kube/secrets/manifest.go
HeavyWombat: Yet another interesting point. I have not thought about it much and basically followed what was defined in the tracker. Now that I re-read the Kubernetes documents, I am curios:
> Labels can be used to select objects and to find collections of objects that satisfy certain conditions

With that statement in mind, from the three labels we define, one should be an annotation I would say. The _sources description_ is the only one that is not serving as an identifier. One could argue that you might want to use it to select all secrets that were created by a specific CRD, but without any real use case I would say this sounds like an edge case scenario. The version number and deployment name however are very good to select objects based on a condition.

HeavyWombat: That being said, if we decide on that set of labels and annotations, I would have to rewrite part of the code to actually get a secret based on the labels and not based on the name.

pkg/kube/secrets/manifest_test.go
HeavyWombat: I think we initially started to arrange the test cases by using the `context` for each list item of the requirements from the tracker card. I am happy to rearrange it based on `describe`.

HeavyWombat:
I updated the PR with another commit that includes changes based on feedback from the review. The change from `kubernetes.Interface` to `controller-runtime` `client.Client` will require a bit of rework and I will do it later.

HeavyWombat:
I made yet another update to the PR: Switched to `controller-runtime` `client.Client`.

pkg/kube/secrets/manifest.go
HeavyWombat: I changed the code with regards to the comments.

pkg/kube/secrets/manifest.go
manno: @viovanov  any opinion, should we invest time here?

viovanov: I think labels are appropriate for versions in this case.
Keep in mind that the Version here is the iteration of the deployment manifest.
It's useful user-facing information, that can be used to identify things (for a diff in scripts for example).

If I understand this statement correctly:
```
Labels allow for efficient queries and watches and are ideal for use in UIs and CLIs. Non-identifying information should be recorded using annotations.
```

Then I think labels are being used correctly, as they are _identifying_ the deployment manifest, as @HeavyWombat 
 mentioned. I also agree with the exception - `sourceDescription`.
Please make that an annotation, then we can move on.

qu1queee:
@viovanov we updated the PR by moving the description into annotations. Can you take a look?

PR #60, opened 2018-12-18
ExtendedJob - Part 1

manno:
Query event log for changed pods and run jobs

* check event log periodically for pod events
* uses a timestamp on the pod to remember when an event triggered a job
* will only run jobs for events more recent than the timestamp
* supports filtering pods on labels

* missing integration test
* only supported job is a sleeping pod for now
* no support for other selectors
* not processing any output yet

[#162413776](https://www.pivotaltracker.com/story/show/162413776)


viovanov:
@manno can you please rebase so I can start my review?

PR #65, opened 2019-01-22
Start doc about bosh to kube conversion

viovanov:


docs/rendering_templates.md
```diff
@@ -0,0 +1,118 @@
...
+## Input & Output
+
+Input is the desired manifest, generated `Secrets` for all referenced variables and information required for discovery of network addresses.
+
+We take advantage of the `ExtendedJob`'s feature to persist output in a `ConfigMap` or `Secret`.
+
+The output of the rendering process is a JSON object that contains all the rendered templates:
+
+The following is and example definition for the `ExtendedJob` that renders templates for `instance-group-2` and `release-a`.
+The release is inferred from the image itself.
```
jandubois: This currently doesn't work; requires https://www.pivotaltracker.com/story/show/163013866 to be finished

docs/rendering_templates.md
```diff
@@ -0,0 +1,118 @@
...
+          command: ["some-ruby-script-that-renders"]
+          env:
+          - name: "INSTANCE_GROUP_NAME"
+            value: "instance-group-2"
+          volumeMounts:
+          - name: deployment-manifest
+            mountPath: /opt/fissile/
+          volumeMounts:
+          - name: mydeployment-bar
+            mountPath: /opt/fissile/
```
jandubois: I would have thought we would do a single mount for *all* the variables, so there should be no `-bar` in the name.

Furthermore I would want the variables in a subdirectory, with nothing but variables in there...

docs/rendering_templates.md
```diff
@@ -0,0 +1,118 @@
...
+        containers:
+        - image: "cfcontainerization/capi-release:opensuse-42.3-24.g63783b3-30.66-1.75.0"
+          command: ["some-ruby-script-that-renders"]
+          env:
+          - name: "INSTANCE_GROUP_NAME"
+            value: "instance-group-2"
+          volumeMounts:
+          - name: deployment-manifest
+            mountPath: /opt/fissile/
+          volumeMounts:
```
jandubois: Delete this line! 😄

docs/rendering_templates.md
```diff
@@ -0,0 +1,118 @@
...
+    spec:
+      template:
+        spec:
+        containers:
+        - image: "cfcontainerization/capi-release:opensuse-42.3-24.g63783b3-30.66-1.75.0"
+          command: ["some-ruby-script-that-renders"]
+          env:
+          - name: "INSTANCE_GROUP_NAME"
+            value: "instance-group-2"
+          volumeMounts:
```
jandubois: Missing one volumemount for the configmap that includes the actual rendering script and support. See https://gist.github.com/jandubois/07c86cde7f23a58dba38d733cf08ea20 for example.

docs/rendering_templates.md
```diff
@@ -0,0 +1,118 @@
...
+            name: "mydeployment-bar"
+            items:
+            - key: "value"
+              path: "bar"
+```
+
+The following is an example output for `instance-group-2` and `release-a`
+```json
+{
+    "job1":{
```
jandubois: Any particular reason to have the release name as a key, instead of including it in the path name, like `job1/bin/start_ctl.sh`?

It provides the additional flexibility to mount the whole secret under the `/var/vcap/jobs` mount point without having to enumerate each file individually.

jandubois: Ok, further research shows that you cannot have slashes in key names; they need to match `[-._a-zA-Z0-9]+`.

This also means that the whole configmap cannot be mounted automatically and maintain the directory structure; that is only possible with PVs afaict.

We will have to specify the mount path for each file individually. On the plus side this means we also get to specify the file mode separately, so only files under `bin/` need to be made executable.

I would suggest that we simply number all rendered files, and then include an extra key that just concatenates the relative paths for each file, e.g.

```
{
    "1": "content of file 1",
    "2": "content of file 2",
    "_": "job1/bin/pre-start:job1/config/my.conf"
}
```

We could maybe use serialized JSON instead of joining the paths with `:`, but I don't care much either way.

jandubois: Actually, this may need more bike-shedding: it will probably be useful to have the mapping for each job separately. At which time using serialized JSON becomes more attractive. So instead of

```
    "_": "job1/bin/pre-start:job1/config/my.conf"
```

we would have

```
    "job1": "{ \"bin/prestart\": \"1\", "config/my.conf\": \"2\" }"
```

with additional entries for each additional job in the release.

PR #73, opened 2019-02-01
Improve docs

qu1queee:
- Fix cf-operator-ci url
- Missleading definition for resources, by using groups ?

I found the usage of `gropus` missleading in the development document. While the `group` is only a part of the resource itself, similar to Version and Kind

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,7 +1,7 @@
 - [Makefile](#makefile)
 - [CI](#ci)
 - [Publishing](#publishing)
-- [Creating a new Group and Controller](#creating-a-new-group-and-controller)
+- [Creating a new Resource and Controller](#creating-a-new-resource-and-controller)
 - [Versioning](#versioning)
 
 ## Makefile
@@ -10,11 +10,11 @@ Before starting, run `make tools` to install the required dependencies.
 
 ## CI
 
-Our Concourse pipeline definitions are kept in the [https://github.com/cloudfoundry-incubator/cf-operator-ci](cf-operator-ci) repo.
+Our Concourse pipeline definitions are kept in the [cf-operator-ci](https://github.com/cloudfoundry-incubator/cf-operator-ci) repo.
 
 ## Publishing
 
-## Creating a new Group and Controller
+## Creating a new Resource and Controller
 
 - create a new directory: `./pkg/kube/apis/<group_name>/<version>`
 - in that directory, create the following files:
@@ -28,12 +28,12 @@ Our Concourse pipeline definitions are kept in the [https://github.com/cloudfoun
   The `types.go` file contains the definition of your resource. This is the file you care about. Make sure to run `make generate` _every time you make a change_. You can also check to see what changes would be done by running `make verify-gen-kube`.
 
   The `register.go` file contains some code that registers your new types.
-  This file looks almost the same for all API groups.
+  This file looks almost the same for all API resources.
 
   The `doc.go` (deep object copy) is required to make the `deepcopy` generator work.
   It's safe to copy this file from another controller.
 
-- in `bin/gen-kube`, add your group to the `GROUP_VERSIONS` variable (separated by a space `" "`):
+- in `bin/gen-kube`, add your resource to the `GROUP_VERSIONS` variable (separated by a space `" "`):
 
   ```bash
   # ...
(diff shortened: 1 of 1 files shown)


manno:
@qu1queee I thought the API group is the top element: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups

And the script we use to generate the new API group for our resource is also named `generate-groups.sh`: https://github.com/cloudfoundry-incubator/cf-operator/blob/5094919239e223c43d14e4e9f0d9c9de210c4b97/bin/gen-kube#L22

Found this:

>  Group: Kubernetes REST APIs are organized in a hierarchy with ‘/apis’ as the root. A group defines a logical name for a set of REST resources under this root. An example API group is ‘apps’, which is represented in the hierarchy as ‘/apis/apps’. You can check all the available API groups using: `kubectl get --raw /apis`  (https://itnext.io/deep-dive-into-how-kubernetes-rest-api-works-517c86f1640b)

qu1queee:
From my understanding, a Kube Resource is a declarative API(`A declarative API expresses a fixed state that the cluster must continually work towards`) with an Schema and Struct(which is the idea behind what we put under `./pkg/kube/apis/<group_name>/<version>`). The resource Schema is composed of the Group/Version/Kind, where the `group`  is used only to avoid issues when having multiple resources with the same `Kind`.(Mainly this statements comes from https://book.kubebuilder.io/basics/what_is_a_resource.html). 

I think this [link](https://kubernetes.io/docs/reference/using-api/api-concepts/#standard-api-terminology) highlight the concept of `groups` as a way to classify different resources, together with he `version`, for an specific resource type.

Does this make sense to you?

qu1queee:
I think the following explains more in detail what I meant, where we have a single group `fissile.cloudfoundry.org` running multiple KUBE API resources:
```
$ ./_output/local/go/bin/kubectl api-resources  --api-group fissile.cloudfoundry.org -o wide
NAME                   SHORTNAMES   APIGROUP                   NAMESPACED   KIND                  VERBS
boshdeployments                     fissile.cloudfoundry.org   true         BOSHDeployment        [delete deletecollection get list patch create update watch]
extendedjobs                        fissile.cloudfoundry.org   true         ExtendedJob           [delete deletecollection get list patch create update watch]
extendedsecrets                     fissile.cloudfoundry.org   true         ExtendedSecret        [delete deletecollection get list patch create update watch]
extendedstatefulsets                fissile.cloudfoundry.org   true         ExtendedStatefulSet   [delete deletecollection get list patch create update watch]
dyn-9-152-98-114:kubernetes encalada$
```

From the above, you can see I use kubectl to list me the KUBE API resources of an specific group, and not groups of a resource. From my interpretation, KIND is the resource name. As I mentioned before, we are building resources for an specific group, not groups.

Opinions?

manno:
The doc for which you opened this PR describe how to run code-gen and create a controller for a new CRD. If I remember correctly we decided on different API groups in `bin/gen-kube` and the filesystem, to *keep the components separate*.

However at the end of the document it says to

>  create a custom resource definition in deploy/helm/cf-operator/templates/

but does not go into detail at all. 

If you run `kubectl api-resources` you're looking at the state in kube, produced by applying our CRDs.
Our code is only tied to that fissile 'group' by `pkg/kube/apis/apis.go`. The CRDs are not generated.

This is all very confusing, we generate the folder structure with a different group version, than the one we use in our `SchemeGroupVersion`: https://github.com/cloudfoundry-incubator/cf-operator/blob/f1aa2423e12d43cf5b9ddd94bb6e5b18c77e1c8f/pkg/kube/apis/extendedjob/v1alpha1/register.go#L22

Well, at least we're missing documentation on how to create new CRDs and how they relate to 

* the generated code (client, lister and deepcopy) and
* our controllers 

So the 'group version' contains the API group, but we're using the same one for all our structs and schemas. This was introduced in 6f06b45 and we created new resources accordingly, that might go back to the beginnings, when we were using the operator-framework to generate controllers.

I'm not sure we want all our 'kinds'/resources in the same 'group version'?
But if we wanted our resources in the same group, a layout like https://github.com/kubernetes/kubernetes/tree/master/pkg/apis/batch would be clearer?

```
% kubectl api-resources  --api-group batch 
NAME       SHORTNAMES   APIGROUP   NAMESPACED   KIND      VERBS
cronjobs   cj           batch      true         CronJob   [create delete deletecollection get list patch update watch]
jobs                    batch      true         Job       [create delete deletecollection get list patch update watch]
```

viovanov:
I believe the word "group" is misused in the `generate-groups.sh` script.

Looking at the [`generate-groups.sh` script](https://github.com/kubernetes/code-generator/blob/master/generate-groups.sh#L31) and [what we pass for that arg](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/bin/gen-kube#L20), it would seem **group** means collection of generated things?

Looking at the Kube docs and the output of `kubectl api-resources`, we clearly have one `APIGroup` as defined [here](https://github.com/cloudfoundry-incubator/cf-operator/blob/ec9e5d601ae76a730e44e554af5eb386c4b0700f/pkg/kube/apis/apis.go#L12), not 4.

So I believe @qu1queee is correct.
@manno if you think there's an inconsistency in the way we generate stuff, please create a ticket.

PR #98, opened 2019-02-13
Create bosh manifest schema and write tests

rohitsakala:
[#163781893](https://www.pivotaltracker.com/story/show/163781893)


pkg/bosh/manifest/manifest.go
```diff
@@ -1,12 +1,202 @@
...
+	VMExtensions       []string               `yaml:"vm_extensions,omitempty"`
+	VMResources        *VMResource            `yaml:"vm_resources"`
+	Stemcell           string                 `yaml:"stemcell"`
+	PersistentDisk     *int                   `yaml:"persistent_disk,omitempty"`
+	PersistentDiskType string                 `yaml:"persistent_disk_type,omitempty"`
+	Networks           []*Network             `yaml:"networks,omitempty"`
+	Update             *Update                `yaml:"update,omitempty"`
+	MigratedFrom       *MigratedFrom          `yaml:"migrated_from,omitempty"`
+	LifeCycle          string                 `yaml:"lifecycle,omitempty"`
+	Properties         map[string]interface{} `yaml:"properties,omitempty"`
```
manno: We need interface{}, because the value could be a string, integer or boolean, right?

pkg/bosh/manifest/manifest.go
```diff
@@ -1,12 +1,202 @@
...
+	RandomizeAzPlacement *bool `yaml:"randomize_az_placement,omitempty"`
+	UseDNSAddresses      *bool `yaml:"use_dns_addresses,omitempty"`
+	UseTmpfsJobConfig    *bool `yaml:"use_tmpfs_job_config,omitempty"`
+}
+
+// AuthType from BOSH deployment manifest
+type AuthType string
+
+const (
+	clientAuth AuthType = "client_auth"
```
manno: If these were public we could use them in other packages instead of the string values?

pkg/bosh/manifest/manifest.go
```diff
@@ -1,12 +1,202 @@
...
+}
+
+// Update from BOSH deployment manifest
+type Update struct {
+	Canaries        int     `yaml:"canaries"`
+	MaxInFlight     string  `yaml:"max_in_flight"`
+	CanaryWatchTime string  `yaml:"canary_watch_time"`
+	UpdateWatchTime string  `yaml:"update_watch_time"`
+	Serial          bool    `yaml:"serial,omitempty"`
+	VMStratergy     *string `yaml:"vm_stratergy,omitempty"`
```
manno: Typo
```suggestion
	VMStrategy     *string `yaml:"vm_strategy,omitempty"`
```

pkg/bosh/manifest/manifest_test.go
```diff
@@ -0,0 +1,409 @@
...
+)
+
+var _ = Describe("Manifest", func() {
+	var (
+		job                 Job
+		jobY                []byte
+		instanceGroup       InstanceGroup
+		instanceGroupY      []byte
+		manifest            Manifest
+		manifestY           []byte
```
manno: Do these have to be global, they're not shared between contexts and not used in `BeforeEach`.

Also  the 'Y' suffix confused me while reading.

pkg/bosh/manifest/manifest_test.go
```diff
@@ -0,0 +1,409 @@
...
+		addOnJobY           []byte
+		err                 error
+	)
+
+	vmResource = VMResource{CPU: 2,
+		RAM:               2048,
+		EphemeralDiskSize: 4096,
+	}
+
+	Context("when marshalling & unmarshalling vm resources schema", func() {
```
manno: Let's talk about go/ginkgo tests when you have time. I think it's enough to check if the struct has the right tags.

Like here: https://github.com/cloudfoundry/bosh-cli/blob/master/cmd/opts_test.go#L99-L105

pkg/bosh/manifest/manifest.go
```diff
@@ -1,12 +1,202 @@
...
 type InstanceGroup struct {
-	Name      string
-	Instances int
+	Name               string                 `yaml:"name"`
+	Instances          int                    `yaml:"instances"`
+	Azs                []string               `yaml:"azs"`
+	Jobs               []Job                  `yaml:"jobs"`
+	VMType             string                 `yaml:"vm_type,omitempty"`
+	VMExtensions       []string               `yaml:"vm_extensions,omitempty"`
+	VMResources        *VMResource            `yaml:"vm_resources"`
```
manno: I'd prefer to avoid  pointers to struct in the data structure. My reasoning is, that we want to avoid shared mutable state, so we don't have side effects. We also want to avoid pointer nil checks.

I don't think performance, memory consumption is  a problem, since we don't change the struct a lot and it's rather small. I wouldn't try to optimize that without benchmarking real world usage first, anyhow.

manno: Ah, I see, we can't have optional fields in our struct unless they are pointers?

rohitsakala: Yes. all fields pop up with default values if we don't use pointers.

pkg/bosh/manifest/manifest_test.go
rohitsakala: Done. I have changed the test cases to test the struct tags only.

rohitsakala:
> Can squash the commits please?

Done.

PR #104, opened 2019-02-14
Update specs

viovanov:


docs/from_bosh_to_kube.md
```diff
@@ -335,6 +345,54 @@ tags:
 
 ### Healthchecks
 
+## Conversion Details
+
+### Calculation of docker image location for releases
+
+### Variables to Extended Secrets
+
+For each **BOSH Variable**, the cf-operator creates `ExtendedSecrets`.
```
jandubois: I don't understand the need to store BOSH variables individually; they could all be stored in a single secret per deployment.

docs/rendering_templates.md
```diff
@@ -94,16 +94,38 @@ Given the information in the **BOSH Job Spec Secrets** and the **Desired Manifes
...
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: api-group-0-cfdeployment
+spec:
+  type: ClusterIP
+  selector:
+    statefulset.kubernetes.io/pod-name: api-group-v1-0
+    statefulset.kubernetes.io/pod-name: api-group-v2-0
```
jandubois: Not sure where the `v1` and `v2` come from.

I assume we have separate ExtendedStatefulSets for each AZ, so I would like to see the AZ name in the pod name as well as in the service name.

docs/from_bosh_to_kube.md
viovanov: They're all created individually using `ExtendedSecrets`.
The size of a large secret could likely go over the limit.

docs/rendering_templates.md
viovanov: This is how `ExtendedStatefulSets` work - `v1` and `v2` are revisions. See [here](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/controllers/extendedstatefulset.md#extended-upgrade-support) for more details.

About AZs - `ExtendedStatefulSets` are AZ-aware - so we'll have a separate `StatefulSet` for each AZ.

PR #113, opened 2019-02-20
Add final state for auto errands

aduffeck:
[#164002768](https://www.pivotaltracker.com/story/show/164002768)


pkg/kube/apis/extendedjob/v1alpha1/types.go
```diff
@@ -14,30 +14,29 @@ import (
 // ExtendedJobSpec defines the desired state of ExtendedJob
 type ExtendedJobSpec struct {
 	Output               *Output                `json:"output,omitempty"`
-	Run                  *Run                   `json:"run,omitempty"`
-	Triggers             *Triggers              `json:"triggers,omitempty"`
+	Trigger              *Trigger               `json:"trigger,omitempty"`
```
manno: If this field is not optional it doesn't need to be a pointer. If it's a pointer the following code is missing nil checks, i.e. `errand_controllger.go:39`

PR #120, opened 2019-02-22
Convert releases to docker image references

aduffeck:
[#164072600](https://www.pivotaltracker.com/story/show/164072600)


jandubois:
I know it is already merged, but this PR introduced a lint error that should be addressed:

```
cf-operator/pkg/bosh/manifest/kube_converter.go:82:1: exported method Manifest.GetReleaseImage should have comment or be unexported
```

PR #147, opened 2019-03-14
IMP Desired Manifest with resolved variables

edwardstudy:
Hi, @viovanov 

As we talked today, variable generation(Variables are replaced by a Job that runs the operator's image) should be done before variable interpolation. So variable interpolation should replace old placeholders in job properties. And the controller can know what kind of variable it needs to mount.

Please review code and leave comments. I'll update code and relative test cases soon.

Best,
Edward


pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -97,6 +116,16 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// TODO Need to update instanceState after finishing Variable Generation stuff
+
+	instanceState = "Variable Interpolation"
```
viovanov: Each state needs to happen asynchronously. Reconcile happens, we look at state - make some action.
When that action is complete, reconcile has to happen again, and we need to update state. Given that we are in a different state, we take a different action. 
Variable interpolation should happen in a job, since you don't have variables generated yet.

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -108,6 +164,200 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+			Name:      manifest.GenerateSecretName(manifest.Name),
+			Namespace: request.Namespace,
+		},
+		StringData: map[string]string{
+			"manifest.yaml": string(tempManifestBytes),
+		},
+	}
+
+	// Check if temp manifest secret already exists
+	foundSecret := &corev1.Secret{}
```
viovanov: This entire section shouldn't be required.
What needs to happen:

1. You use the kube converter function (https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/bosh/manifest/kube_converter.go#L56) to generate `ExtendedSecret` definitions
2. You loop through them and create them
3. You create an `ExtendedJob` that uses the secrets *that will be created by the ExtendedSecrets*

You don't need to have logic that tests if the secrets have been created.
You don't need to create the secret that contains the interpolated manifest, the extended job does that on its own.

edwardstudy:
Hi, @viovanov. I updated code and implemented variable generation part. Please help to review. 

Thanks!

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -83,50 +128,345 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+		labelManifestSHA1: currentManifestSHA1,
+	}
+
+	defer func() {
+		// Fetch latest BOSHDeployment before update
+		foundInstance := &bdc.BOSHDeployment{}
+		key := types.NamespacedName{Namespace: instance.GetNamespace(), Name: instance.GetName()}
+		err := r.client.Get(ctx, key, foundInstance)
+		if err != nil {
+			r.log.Errorf("Failed to get BOSHDeployment instance '%s': %v", instance.GetName(), err)
```
viovanov: Shouldn't we return at this point?

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -83,50 +128,345 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+		labelDeployment: manifest.Name,
+	}
+	desiredManifestSecretLabels := map[string]string{
+		labelKind:         "desired-manifest",
+		labelDeployment:   manifest.Name,
+		labelManifestSHA1: currentManifestSHA1,
+	}
+
+	defer func() {
+		// Fetch latest BOSHDeployment before update
```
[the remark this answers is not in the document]
viovanov: Also, please explain why this needs to happen in `defer`

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -83,50 +128,345 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+			Namespace: instance.GetNamespace(),
+		},
+		StringData: map[string]string{
+			"manifest.yaml": string(tempManifestBytes),
+		},
+	}
+
+	foundSecret := &corev1.Secret{}
+	err = r.client.Get(ctx, types.NamespacedName{Name: tempManifestSecret.GetName(), Namespace: tempManifestSecret.GetNamespace()}, foundSecret)
+	if apierrors.IsNotFound(err) {
```
viovanov: can you please explain what this `tempManifestSecret` is used for?

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -83,50 +128,345 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+		}
+		err = r.client.Update(ctx, foundSecret)
+		if err != nil {
+			return errors.Wrap(err, "could not update temp manifest secret")
+		}
+	}
+
+	r.log.Debug("Creating variable interpolation extendedJob")
+	varIntExJob, err := r.newExtendedJobTemplateForVariableInterpolation(ctx, foundSecret, variables, jobLabels, secretLabels, instance.GetNamespace())
+	if err != nil && apierrors.IsNotFound(err) {
```
viovanov: Why do you check for `IsNotFound`?
Even if the secrets have not yet been created, afaik you can still create the job. It will keep trying to run until all secrets are ready.

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -135,24 +475,299 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+func (r *ReconcileBOSHDeployment) actionOnDeploying(ctx context.Context, instance *bdc.BOSHDeployment, kubeConfigs *bdm.KubeConfig) error {
+	// TODO Check deployment
+	instance.Status.State = deployedState
+
+	return nil
+}
+
+// newExtendedJobTemplateForVariableInterpolation returns a job to interpolate variables
+func (r *ReconcileBOSHDeployment) newExtendedJobTemplateForVariableInterpolation(ctx context.Context, manifest *corev1.Secret, variables []esv1.ExtendedSecret, jobLabels map[string]string, secretLabels map[string]string, namespace string) (*ejv1.ExtendedJob, error) {
+	cmd := []string{"/bin/sh"}
```
viovanov: I think you can run the `cf-operator` directly - no need for a shell.

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -135,24 +475,299 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+	args := []string{"-c", "cf-operator variable-interpolation --manifest /var/run/secrets/manifest/manifest.yaml --variables-dir /var/run/secrets/variables | base64 | tr -d '\n' | echo \"{\\\"interpolated-manifest.yaml\\\":\\\"$(</dev/stdin)\\\"}\""}
```
viovanov: Try using back-ticks - 
``` 
`this is "a string with double quotes that don't need escaping"`
```

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -135,24 +475,299 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+						},
+					},
+				},
+			},
+		},
+	}
+	volumeMounts := []corev1.VolumeMount{
+		{
+			Name:      string(manifest.GetUID()),
+			MountPath: "/var/run/secrets/manifest",
```
viovanov: No need for the double "manifest" - `manifest/manifest.yaml`.
Can you mount directly to `/var/run/secrets`?

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -135,24 +475,299 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+		},
+	}
+
+	for _, variable := range variables {
+		vars := Variables{}
+		varName := variable.GetLabels()["variableName"]
+
+		secretItems := []corev1.KeyToPath{}
+
+		// Check if variable already exists
```
viovanov: Don't check.
Just create the ejob - it will **eventually** run, once everything is generated.

pkg/kube/controllers/boshdeployment/reconciler.go
```diff
@@ -135,24 +475,299 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
...
+		if err != nil && apierrors.IsNotFound(err) {
+			r.log.Debugf("Could not find secret '%s'", variable.GetName())
+			return nil, err
+		} else if err != nil {
+			r.log.Errorf("Failed to get secret '%s': %v", variable.GetName(), err)
+			return nil, err
+		}
+
+		// Update secret and relative SecretVolumeSource
+		switch variable.Spec.Type {
```
viovanov: I don't understand this switch, and why we update all the secrets.
Can you please explain?

viovanov: It looks like we're moving from `data` to `stringData` - this shouldn't be required.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: I think I should explain what func `newExtendedJobTemplateForVariableInterpolation` does:
For variable interpolation job, this job should get variable values as this structure: https://bosh.io/docs/variable-types

For example, this interpolated ssh var can be used as `diego_ssh_proxy_host_key.private_key`:
**structure 1**
```
diego_ssh_proxy_host_key:
  private_key: |
    *****
  public_key: |
    *****
  public_key_fingerprint: *****
```

And our extendSecret just generates a plain structure:
**structure 2**
```
data:
  SSHPrivateKey: ***
  SSHPublicKey: ***
  SSHFingerprint: ***
```
And that's why this func constructs a `***-gen` key to each variable and move **structure 2** to **structure 1**. From my point of view, replacing existing variable secrets just works. But we can discuss a better solution.

And for this reason, this func should wait for secret to be created and replace it.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: Reply as above comments: https://github.com/cloudfoundry-incubator/cf-operator/pull/147#discussion_r268016548

pkg/kube/controllers/boshdeployment/reconciler.go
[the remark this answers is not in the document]
edwardstudy: What this defer wanna do is updating instance state if any error occurs.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: As interpolation job input, this manifest has been already applied ops files.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: This snippet is to upsert temp manifest secret. So it will check `IsNotFound ` and then create or update secret.

pkg/kube/controllers/boshdeployment/reconciler.go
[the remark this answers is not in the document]
viovanov: Ok, please make this a separate function and then
```
defer reconcileDeploymentManifestState(instance)
```

pkg/kube/controllers/boshdeployment/reconciler.go
viovanov: 👍 
Makes sense - please add a comment with that information

pkg/kube/controllers/boshdeployment/reconciler.go
viovanov: No, afaict `newExtendedJobTemplateForVariableInterpolation` returns errors if variable secrets don't exist.
See here: https://github.com/cloudfoundry-incubator/cf-operator/pull/147/files/7a9e90535b997b218f8e7ee1f1152ba3b545595c#diff-3c2f267f825b15e51537a7f733f474deR550

pkg/kube/controllers/boshdeployment/reconciler.go
viovanov: Ok, so this is just about how the data looks.
We do not want to process data here. 
We have 2 options:

1. Change the way `ExtendedSecrets` place data in the secrets
2. Mount the data as is, and move data processing to the actual interpolation command

Because our API is not stable, I'd like us to do option 1.
Another reason is that I don't like the usage of Capital Leters in those keys (they can become filenames when mounting the secret).

pkg/kube/controllers/boshdeployment/reconciler.go
viovanov: Yep, thanks for the explanation.
Let's change extended secrets so they create what we need, and then remove all of this.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: experiment is failed.

pkg/kube/controllers/boshdeployment/reconciler.go
edwardstudy: Sorry. I just mis-replied above section. Yes. What this section want to do is keep trying to run until all secrets are ready. When one secret didn't be created, this func will return `IsNotFound` error to here: https://github.com/cloudfoundry-incubator/cf-operator/blob/76f21dd108fd7ca80ced0ee2115b75e8d390219c/pkg/kube/controllers/boshdeployment/reconciler.go#L199-L203

And reconciler will retry this action.

PR #149, opened 2019-03-18
data gatherer algorithm implementation

qu1queee:
Translate data-gathering algorithm from ruby to golang

- Generate Job Specs and Links based on providers
- Ensure consumer jobs can see provided links information
- Able to render bpm files, with correct referenced parameters
  which return desired values
- Parse the bpm rendered file into the appropiate bpm struct
- Generate final manifest in JSON format
- Add test cases that cover mainly the struct under the job.
  properties.bosh_containerization:
  - consumers
  - instances
  - providers
  - instances.BPM
  - links
- Add sample *.MF files
- Define helper functions that are require for:
  - convert map[interface{}]interface{} to map[string]interface{}
  - lookup releases on manifest
  - generated nested structs based on strings like foo.bar
- Ensure shared bpm instance is stored as top level property.


qu1queee:
Current PR fails due to ruby missing on the `cfcontainerization/go-tools`. I just open a PR on the ci repo https://github.com/cloudfoundry-incubator/cf-operator-ci/pull/18 for it.

PR #151, opened 2019-03-19
Run instance groups on deployment manifest creation and fix tests

viovanov:
[#164072732](https://www.pivotaltracker.com/story/show/164072732)

### DIFF
--- a/Dockerfile-nobuild
+++ b/Dockerfile-nobuild
@@ -1,0 +1,4 @@
+FROM opensuse/leap:15.0
+RUN zypper -n in system-user-nobody
+USER nobody
+COPY binaries/cf-operator /usr/local/bin/cf-operator
--- a/bin/up
+++ b/bin/up
@@ -9,6 +9,12 @@ if type -a minikube >/dev/null 2>/dev/null; then
     eval ${dockerenv}
 fi
 
+GIT_ROOT="${GIT_ROOT:-$(git rev-parse --show-toplevel)}"
+. "${GIT_ROOT}/bin/include/versioning"
+. "${GIT_ROOT}/.envrc"
+
+export DOCKER_IMAGE_TAG=${VERSION_TAG}
+
 set -e
 
 bin/build-image
--- a/integration/deploy_test.go
+++ b/integration/deploy_test.go
@@ -3,26 +3,28 @@ package integration_test
 import (
 	"time"
 
+	"code.cloudfoundry.org/cf-operator/integration/environment"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Deploy", func() {
 	Context("when correctly setup", func() {
-		podName := "diego-pod"
+		podName := "my-manifest-nats-v1-0"
+		stsName := "my-manifest-nats-v1"
 
 		AfterEach(func() {
-			env.WaitForPodsDelete(env.Namespace)
+			Expect(env.WaitForPodsDelete(env.Namespace)).To(Succeed())
 		})
 
 		It("should deploy a pod with 10 seconds for the reconciler context", func() {
 			tearDown, err := env.CreateConfigMap(env.Namespace, env.DefaultBOSHManifestConfigMap("manifest"))
 			Expect(err).NotTo(HaveOccurred())
-			defer tearDown()
+			defer func(tdf environment.TearDownFunc) { Expect(tdf()).To(Succeed()) }(tearDown)
 
 			_, tearDown, err = env.CreateBOSHDeployment(env.Namespace, env.DefaultBOSHDeployment("test", "manifest"))
 			Expect(err).NotTo(HaveOccurred())
-			defer tearDown()
+			defer func(tdf environment.TearDownFunc) { Expect(tdf()).To(Succeed()) }(tearDown)
 
 			// check for pod
 			err = env.WaitForPod(env.Namespace, podName)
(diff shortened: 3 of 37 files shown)


pkg/kube/controllers/extendedjob/errand_reconciler.go
```diff
@@ -107,6 +107,11 @@ func (r *ErrandReconciler) Reconcile(request reconcile.Request) (result reconcil
 }
 
 func (r *ErrandReconciler) createJob(ctx context.Context, extJob ejv1.ExtendedJob) error {
+	if extJob.Spec.Template.Labels == nil {
+		extJob.Spec.Template.Labels = map[string]string{}
+	}
+	extJob.Spec.Template.Labels["ejob-name"] = extJob.Name
```
manno: I don't understand this code. It modifies the extended job and relies on later code to store this?
Are labels transferred from the extended job to the pod somehow, or how does clean up code find pods with that label?

Could this be moved to be near the actual `Update(extjob)` statement, it's not  obvious from naming and function signature, that the passed in `extJob` is mutated here.

pkg/kube/controllers/extendedjob/trigger_reconciler.go
```diff
@@ -127,6 +127,11 @@ func (r *TriggerReconciler) Reconcile(request reconcile.Request) (result reconci
 }
 
 func (r *TriggerReconciler) createJob(ctx context.Context, extJob ejv1.ExtendedJob, podName string) error {
+	if extJob.Spec.Template.Labels == nil {
+		extJob.Spec.Template.Labels = map[string]string{}
+	}
+	extJob.Spec.Template.Labels["ejob-name"] = extJob.Name
```
manno: Same as errand reconciler comment

integration/environment/environment.go
```diff
@@ -52,7 +55,7 @@ func NewEnvironment() *Environment {
 			Fs:         afero.NewOsFs(),
 		},
 		Machine: Machine{
-			pollTimeout:  30 * time.Second,
+			pollTimeout:  60 * time.Second,
```
manno: Is it necessary to increase the timeout?

integration/extended_job_test.go
```diff
@@ -194,33 +195,34 @@ var _ = Describe("ExtendedJob", func() {
...
 			tearDown, err = env.CreatePod(env.Namespace, pod)
 			Expect(err).NotTo(HaveOccurred())
-			defer tearDown()
+			defer func(tdf environment.TearDownFunc) { Expect(tdf()).To(Succeed()) }(tearDown)
 
 			By("waiting for the job")
 			_, err = env.CollectJobs(env.Namespace, "extendedjob=true", 1)
 			Expect(err).NotTo(HaveOccurred(), "error waiting for jobs from extendedjob")
 
 			// Why does the ExtendedJob deletion not trigger the job deletion?
```
manno: This comment can probably be removed now?

testing/catalog.go
```diff
@@ -572,9 +692,8 @@ func (c *Catalog) AnnotatedPod(name string, annotations map[string]string) corev
 
 // Sleep1hPodSpec defines a simple pod that sleeps 60*60s for testing
 func (c *Catalog) Sleep1hPodSpec() corev1.PodSpec {
-	one := int64(1)
 	return corev1.PodSpec{
-		TerminationGracePeriodSeconds: &one,
+		TerminationGracePeriodSeconds: helper.Int64(1),
```
manno: We should replace all occurrences with the helper, not just this one.

integration/deploy_test.go
```diff
@@ -3,26 +3,28 @@ package integration_test
...
 		AfterEach(func() {
-			env.WaitForPodsDelete(env.Namespace)
+			Expect(env.WaitForPodsDelete(env.Namespace)).To(Succeed())
 		})
 
 		It("should deploy a pod with 10 seconds for the reconciler context", func() {
 			tearDown, err := env.CreateConfigMap(env.Namespace, env.DefaultBOSHManifestConfigMap("manifest"))
 			Expect(err).NotTo(HaveOccurred())
-			defer tearDown()
+			defer func(tdf environment.TearDownFunc) { Expect(tdf()).To(Succeed()) }(tearDown)
```
manno:
```suggestion
			defer func(f environment.TearDownFunc) { Expect(f()).To(Succeed()) }(tearDown)
```
I think just `f` would be just fine here.

Dockerfile-nobuild
```diff
@@ -0,0 +1,4 @@
+FROM opensuse/leap:15.0
+RUN zypper -n in system-user-nobody
+USER nobody
+COPY binaries/cf-operator /usr/local/bin/cf-operator
```
manno: I guess it's not possible to do this in the same dockerfile.  `RUN if` maybe?

pkg/kube/controllers/extendedjob/errand_reconciler.go
viovanov: About moving - I agree about not wanting to mutate it here.
About the other comments, I don't know what you mean. The label will exist on the pod. That's it - it's required to link it back to the ejob. `job-name` is insufficient.

integration/environment/environment.go
viovanov: nope, forgot to switch it back 👍

integration/extended_job_test.go
viovanov: 👍

testing/catalog.go
viovanov: 👍

Dockerfile-nobuild
viovanov: I don't think what you mean is possible.

pkg/kube/controllers/extendedjob/errand_reconciler.go
manno: Yes I missed that the label is set on `Template` which is the actual `PodSpec` being for the pod.

viovanov: 👍

testing/catalog.go
viovanov: https://www.pivotaltracker.com/story/show/164729001

pkg/kube/controllers/extendedjob/errand_reconciler.go
manno: Ah, I see now, we don't actually need to mutate extendjob (not even sure if it does). We just need to pass this modified PodTemplateSpec to Job a few lines below

bin/up
```diff
@@ -9,6 +9,12 @@ if type -a minikube >/dev/null 2>/dev/null; then
     eval ${dockerenv}
 fi
 
+GIT_ROOT="${GIT_ROOT:-$(git rev-parse --show-toplevel)}"
```
manno: I just noticed this introduces `GIT_ROOT` to have the absolute path available, but most of the script still uses relative paths (line 20, ...).
We could remove `GIT_ROOT` here and in `bin/test-integration`, too.

PR #166, opened 2019-03-26
Support self-signed CA variables

edwardstudy:


integration/environment/machine.go
```diff
@@ -45,7 +45,7 @@ func (m *Machine) CreatePod(namespace string, pod corev1.Pod) (TearDownFunc, err
 	client := m.Clientset.CoreV1().Pods(namespace)
 	_, err := client.Create(&pod)
 	return func() error {
-		err := client.Delete(pod.GetName(), &metav1.DeleteOptions{})
+		err := client.Delete(pod.GetName(), metav1.NewDeleteOptions(0))
```
viovanov: Why do you think `grace period = 0` is required?

edwardstudy: I found many integartion-test cases failed due to `timed out waiting for the condition` in and found that pod stuck in `Terminating`.

edwardstudy: OK. After #165 merged, will remove this commit.

aduffeck: #165 is now merged

edwardstudy: Done

docs/controllers/extendedsecret.md
```diff
@@ -26,18 +32,34 @@ The developer can specify policies for rotation (e.g. automatic or not) and how
...
-    # the specific contents for each type of request need to be determined by the definitions here:
-    # https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/credsgen/generator.go#L43
-    # in the case where we need to reference a CA, that should be done by referencing a Secret
-    password:
-    certificate:
-    sshKey:
-    rsaKey:
-```
\ No newline at end of file
+  spec:
```
aduffeck: This is a bit garbled, I think. I believe it should look like this:
```yaml
---
apiVersion: fissile.cloudfoundry.org/v1alpha1
kind: ExtendedSecret
metadata:
  name: my-generated-secret
spec:
  # Type of the variable that ExtendedSecret supports
  type: certificate
  # Name of the Secret that stores this variable
  secretName: cf-deployment.uaa-ssl
  policy:
    recreateInterval: 3600s
  request:
    certificate:
      # The secret of CA private key
      CAKeyRef:
        Key: private_key
        Name: cf-deployment.uaa-ca
      # The secret of CA certificate
      CARef:
        Key: certificate
        Name: cf-deployment.uaa-ca
      # The Subject name of the certificate
      commonName: uaa.service.cf.internal
      # The subject alternative names
      alternativeNames:
      - uaa.service.cf.internal
      # If true, the ExtendedSecret will generate self-signed root CA certificate and private key
      isCA: false
```

pkg/credsgen/in_memory_generator/certificate_test.go
```diff
@@ -41,7 +41,7 @@ var _ = Describe("InMemoryGenerator", func() {
 			)
 
 			BeforeEach(func() {
-				ca, _ = generator.GenerateCertificate("testca", credsgen.CertificateGenerationRequest{IsCA: true})
+				ca, _ = generator.GenerateCertificate("testca", credsgen.CertificateGenerationRequest{CommonName: "Fake CA",IsCA: true})
```
aduffeck: Could we add a test that asserts that the given `CommonName` is actually used for CAs? I think this could go to the test in line 122.

edwardstudy: Done. Add asserts about CN.

PR #170, opened 2019-03-26
Autodetect az,ip,index

rohitsakala:
[#164892003](https://www.pivotaltracker.com/n/projects/2192232/stories/164892003)

### DIFF
--- a/cmd/internal/trender.go
+++ b/cmd/internal/trender.go
@@ -1,9 +1,11 @@
 package cmd
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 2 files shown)


pkg/kube/controllers/extendedstatefulset/pod_mutator.go
```diff
@@ -192,6 +192,13 @@ func replaceVersionInName(name string, version string, offset int) string {
 	return name
 }
 
+// getPodIndex returns ordinal of the Pod
+func getPodIndex(name string) string {
```
aduffeck: This function isn't used anywhere, is it?

cmd/internal/trender.go
```diff
@@ -64,26 +89,28 @@ func init() {
...
 	viper.BindPFlag("address", templateRenderCmd.Flags().Lookup("address"))
 	viper.BindPFlag("az", templateRenderCmd.Flags().Lookup("az"))
 	viper.BindPFlag("id", templateRenderCmd.Flags().Lookup("id"))
 	viper.BindPFlag("index", templateRenderCmd.Flags().Lookup("index"))
 	viper.BindPFlag("name", templateRenderCmd.Flags().Lookup("name"))
 	viper.BindPFlag("ip", templateRenderCmd.Flags().Lookup("ip"))
+
+	viper.BindEnv("az", "CF_OPERATOR_AZ")
+	viper.BindEnv("ip", "POD_IP")
+	viper.BindEnv("azi", "CF_OPERATOR_AZ_INDEX")
```
aduffeck: I don't think this binding is used anywhere.

cmd/internal/trender.go
```diff
@@ -28,9 +30,27 @@ This will render a provided manifest instance-group
...
+			// calculate index
+			azIndex, err := strconv.Atoi(os.Getenv("CF_OPERATOR_AZ_INDEX"))
+			if err != nil {
+				return err
+			}
+			podIndex, err := strconv.Atoi(os.Getenv("POD_INDEX"))
+			if err != nil {
+				return fmt.Errorf("podIndex not found")
+			}
+			index = strconv.Itoa(podIndex * azIndex)
```
aduffeck: Where does this index calculation algorithm come from? There will be conflicts, e.g. [azIndex: 0, podIndex: 1] and [azIndex: 1, podIndex: 0] would both become index 0. Is that a problem?

rohitsakala: @aduffeck Yes. I didn't think of this. This is stated in [Link](https://github.com/cloudfoundry-incubator/cf-operator/blob/58671dbe9a7e910d859f46c17042f89fed054032/docs/rendering_templates.md).  @viovanov can answer maybe.

cmd/internal/trender.go
rohitsakala: Yes. I forgot to remove this.

pkg/kube/controllers/extendedstatefulset/pod_mutator.go
rohitsakala: Yes, This also need to be removed.

PR #177, opened 2019-03-29
Use BPM information when starting instance groups

viovanov:


viovanov:
closed in favor of #179

PR #180, opened 2019-03-29
Use context to store logger

manno:
* introduce ctx as first arg, same parent ctx everywhere
* config now in separate package
* NewManagerContext and NewReconcilerContext
* zap logger in context, removed from func args and structs
* use context.WithTimeout to get warnings for missing cancel() calls
* reorder imports (go, github, k8s, cfo)
* removed context type alias
* context no longer part of any struct, except Reconciler

[#164933036](https://www.pivotaltracker.com/story/show/164933036)

Another PR to add events via recorder will have to follow later.


qu1queee:
@manno a quick question on how to use the context with value. I was of the impression that is not the best practice to store pointers as the values, not sure what is the exact use case here, but I think it´s valid to think about the type of data to store and what to not.

manno:
@qu1queee hm, I was following https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/zap/ctxzap/context.go

I didn't' see any docs discouraging storing the pointer?

qu1queee:
Yes, I think the ctx value should be of the type request scoped data, which means all data that originates in a request and its deleted when the request is over.(https://golang.org/pkg/context/, https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation) , therefore the logger could be not ideal mainly because you lose:
- type safety
- refactoring could be hard later

But I also read this one(https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) , I like the explanation about informing/controlling. As long as the value in the ctx its only used to inform, meaning it will not affect the request somehow, I will say it is fine to keep the logger.

PR #182, opened 2019-04-01
Fix unhandled errors

edwardstudy:


pkg/kube/controllers/webhook_configuration.go
```diff
@@ -144,9 +119,34 @@ func (f *WebhookConfig) setupCertificate() error {
 		f.CaCertificate = caCert.Certificate
 		f.Key = cert.PrivateKey
 		f.Certificate = cert.Certificate
+	} else {
```
viovanov: Do we need an else here? 
You should be able to make sure you have a `data` object in both scenarios.

```
data = data from secret

if secret doesn't exist
  data = generated data
end

set values on f
write to file
```

pkg/kube/controllers/extendedjob/trigger_reconciler.go
```diff
@@ -157,17 +163,28 @@ func (r *TriggerReconciler) createJob(ctx context.Context, extJob ejv1.ExtendedJ
 // jobName returns a unique, short name for a given extJob, pod combination
 // k8s allows 63 chars, but the pod will have -\d{6} appended
 // IDEA: maybe use pod.Uid instead of rand
-func jobName(extJobName, podName string) string {
-	hashID := randSuffix(fmt.Sprintf("%s-%s", extJobName, podName))
-	return fmt.Sprintf("job-%s-%s-%s", truncate(extJobName, 15), truncate(podName, 15), hashID)
+func jobName(extJobName, podName string) (string, error) {
```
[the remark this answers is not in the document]
edwardstudy: done

pkg/kube/controllers/webhook_configuration.go
edwardstudy: data from secret is map[string]interface, but generated  data is map[string][]uint8. So we use if-else to handle this different.

PR #189, opened 2019-04-08
Qu1queee/165014546 bpm per job

qu1queee:
Modify job BPM rendering:
- Avoid adding a BPM object per job instance
- Generate a single BPM object per job
- Validate if all job instances have the same BPM object. This will be
  processed for validation, but at the end, the job instance will not
  stick with a BPM object.
- Refactor test cases for BPM rendering.
- Work some pending TODO´s.

@viovanov could you take a look on this, if this is the intended behaviour, I will then add another commit to modify the docs.

### DIFF
--- a/cmd/internal/dgather.go
+++ b/cmd/internal/dgather.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	btg "github.com/viovanov/bosh-template-go"
+	"go.uber.org/zap"
 	yaml "gopkg.in/yaml.v2"
 
 	"code.cloudfoundry.org/cf-operator/pkg/bosh/bpm"
@@ -32,6 +33,7 @@ inside a bosh manifest file.
 
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		defer log.Sync()
 		boshManifestPath := viper.GetString("bosh-manifest-path")
 		if len(boshManifestPath) == 0 {
 			return fmt.Errorf("manifest cannot be empty")
@@ -84,6 +86,7 @@ inside a bosh manifest file.
 }
 
 func init() {
+	initConfig()
 	rootCmd.AddCommand(dataGatherCmd)
 
 	dataGatherCmd.Flags().StringP("bosh-manifest-path", "m", "", "path to a bosh manifest file")
@@ -287,7 +290,7 @@ func GenerateJobConsumersData(currentJob *manifest.Job, jobReleaseSpecs map[stri
 }
 
 // RenderJobBPM per job and add its value to the jobInstances.BPM field
-func RenderJobBPM(currentJob *manifest.Job, jobInstances []manifest.JobInstance, baseDir string, manifestName string) error {
+func RenderJobBPM(currentJob *manifest.Job, jobInstances []manifest.JobInstance, baseDir string, manifestName string, log *zap.SugaredLogger) error {
 
 	// Location of the current job job.MF file
 	jobSpecFile := filepath.Join(baseDir, "jobs-src", currentJob.Release, currentJob.Name, "job.MF")
@@ -329,7 +332,7 @@ func RenderJobBPM(currentJob *manifest.Job, jobInstances []manifest.JobInstance,
 	jobIndexBPM := make([]bpm.Config, len(jobInstances))
 
 	if jobInstances != nil {
-		for i := range jobInstances {
+		for i, jobInstance := range jobInstances {
 
 			properties := currentJob.Properties.ToMap()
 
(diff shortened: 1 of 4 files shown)


cmd/internal/dgather.go
```diff
@@ -332,19 +333,20 @@ func RenderJobBPM(currentJob manifest.Job, jobInstances []manifest.JobInstance,
...
 
 				&btg.InstanceInfo{
-					Address:    instance.Address,
-					AZ:         instance.AZ,
-					ID:         instance.ID,
-					Index:      string(instance.Index),
+					Address:    jobInstances[0].Address,
+					AZ:         jobInstances[0].AZ,
+					ID:         jobInstances[0].ID,
+					Index:      string(jobInstances[0].Index),
```
viovanov: The point for this loop is to make sure `bpm.yml` doesn't change based on index, address or AZ.
So we need to render bpm with the parameters for each index, to have a correct comparison.

cmd/internal/dgather.go
```diff
@@ -361,18 +363,50 @@ func RenderJobBPM(currentJob manifest.Job, jobInstances []manifest.JobInstance,
...
 
 			// Parse a rendered bpm.yml into the bpm Config struct
-			jobInstances[i].BPM, err = bpm.NewConfig(bpmBytes)
+			jobIndexBPM[i], err = bpm.NewConfig(bpmBytes)
 			if err != nil {
 				return err
 			}
+			instanceBytes := generateJobInstanceSHA(currentJob, jobInstances[i])
+
+			// All instances SHA are the same, while all instances are based on index 0
```
viovanov: I don't understand this comment, can you please explain why we need the SHA, and what do you mean that all instances are based on index 0?

cmd/internal/dgather.go
```diff
@@ -361,18 +363,50 @@ func RenderJobBPM(currentJob manifest.Job, jobInstances []manifest.JobInstance,
...
 
-			// Consider adding a Fingerprint to each job instance
-			// instance.Fingerprint = generateSHA(fingerPrintBytes)
+		for _, jobBPMInstance := range jobIndexBPM {
+			if !reflect.DeepEqual(jobBPMInstance, jobIndexBPM[0]) {
+				return errors.New("found different BPM job indexes, this is not supported")
+			}
 		}
+
+		currentJob.Properties.BOSHContainerization.BPM = jobIndexBPM[0]
```
viovanov: 👍

cmd/internal/dgather.go
```diff
@@ -361,18 +363,50 @@ func RenderJobBPM(currentJob manifest.Job, jobInstances []manifest.JobInstance,
...
 	}
 	return nil
 }
 
+func generateJobInstanceSHA(currentJob *manifest.Job, jobInstance manifest.JobInstance) []byte {
+	// The following is based on https://github.com/cloudfoundry-incubator/cf-operator/compare/master...jandubois:161745895-erb-rendering-using-jobs#diff-55ce7ec0a607742ace583aad21b7acbaR231
+	// Calculate a fingerprint of all config settings for this job instance
```
viovanov: Please add the "why" - would be good as a comment to the function.

cmd/internal/dgather.go
qu1queee: This is related to the previous comment, where I hardcoded the bpm rendering to use always the info of the job instance with index 0.  This will end with all job instances that are passed to bpm being the same, which make the fingerprint for each instance equal. This did not make a lot of sense to me, while to whole idea of sha´s is to be unique per instance.

cmd/internal/dgather.go
qu1queee: This will be addressed on the next commit.

cmd/internal/dgather.go
viovanov: But the fingerprint is not used anywhere afaict.
Why keep it, for what purpose?

cmd/internal/dgather.go
```diff
@@ -282,7 +285,7 @@ func GenerateJobConsumersData(currentJob *manifest.Job, jobReleaseSpecs map[stri
 }
 
 // RenderJobBPM per job and add its value to the jobInstances.BPM field
-func RenderJobBPM(currentJob manifest.Job, jobInstances []manifest.JobInstance, baseDir string, manifestName string) error {
+func RenderJobBPM(currentJob *manifest.Job, jobInstances []manifest.JobInstance, baseDir string, manifestName string, log *zap.SugaredLogger) error {
```
viovanov: Don't pass a `log`, use a `Context`
See here:  https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/util/ctxlog/context.go

Every function should have a context, and the `log` object is retrieved using `ExtractLogger`

cmd/internal/dgather.go
```diff
@@ -361,20 +367,55 @@ func RenderJobBPM(currentJob manifest.Job, jobInstances []manifest.JobInstance,
...
+
+			// All instances SHA are the same, while all instances are based on index 0
+			jobInstances[i].Fingerprint = generateSHA(instanceBytes)
+		}
 
-			// Consider adding a Fingerprint to each job instance
-			// instance.Fingerprint = generateSHA(fingerPrintBytes)
+		for _, jobBPMInstance := range jobIndexBPM {
+			if !reflect.DeepEqual(jobBPMInstance, jobIndexBPM[0]) {
+				log.Infof("found different BPM job indexes for job %v in manifest %v, this is NOT SUPPORTED", currentJob.Name, manifestName)
```
viovanov: add a `WARNING: ` in front of it please.

qu1queee: done

cmd/internal/dgather.go
qu1queee: done

cmd/internal/dgather.go
qu1queee: This was based on Jan´s algorithm, because this is not longer needed, I just removed all logic related to the fingerprint generation.

manno:
Talked to Vlad, rebased against master and removed ctxlog. Backed up old branch to `qu1queee/165014546_bpm_per_job-bak`

PR #190, opened 2019-04-09
Variable secrets labelled `generated` can be overwritten

edwardstudy:
From the description, ExtendedSecert reconciler should run as below:

* Check if the secret was already created and then check if this secret could be generated
  * If the secret  was created **and** could not be generated,  reconciler skip reconciling
  * If the secret  was not created **or** could be generated,  reconciler would create/overwrite a new secret

[#165054299](https://www.pivotaltracker.com/story/show/165054299)

Thanks

### DIFF
--- a/pkg/kube/apis/extendedsecret/v1alpha1/types.go
+++ b/pkg/kube/apis/extendedsecret/v1alpha1/types.go
@@ -1,7 +1,11 @@
 package v1alpha1
 
 import (
+	"fmt"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"code.cloudfoundry.org/cf-operator/pkg/kube/apis"
 )
 
 // This file is safe to edit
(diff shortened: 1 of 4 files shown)


pkg/kube/controllers/extendedsecret/reconciler.go
```diff
@@ -73,50 +73,41 @@ func (r *ReconcileExtendedSecret) Reconcile(request reconcile.Request) (reconcil
...
+	canBeGenerated, err := r.canBeGenerated(ctx, instance)
+	if err != nil {
+		ctxlog.Errorf(ctx, "Error reading the secret: %v", err.Error())
+		return reconcile.Result{}, err
 	}
-	err = r.client.Get(ctx, namespacedName, generatedSecret)
-	if err == nil {
-		ctxlog.Info(ctx, "Skip reconcile: secret already exists")
+	if !canBeGenerated {
+		ctxlog.Debugf(ctx, "SKip reconcile: secret '%s' could not be generated", instance.Spec.SecretName)
```
viovanov: typo in `SKip`.
Also please add more details in the log message, and make it `Info`.
`Skip reconcile: secret '%s' already exists and it's not generated`

pkg/kube/apis/extendedsecret/v1alpha1/types.go
```diff
@@ -19,6 +23,11 @@ const (
 	RSAKey      Type = "rsa"
 )
 
+var (
+	// LabelKind is the label key for secret kind
+	LabelKind = fmt.Sprintf("%s/kind", apis.GroupName)
```
viovanov: please make this `fmt.Sprintf("%s/secret-kind", apis.GroupName)`

PR #192, opened 2019-04-10
Simplify code for data gathering and rendering, by:

qu1queee:
- Appending to initContainers via a single method
- Remove duplication of code
- Add missing test cases for the initContainers of the data
  gathering kubeConverter job

### DIFF
--- a/pkg/bosh/manifest/kube_converter.go
+++ b/pkg/bosh/manifest/kube_converter.go
@@ -295,7 +295,7 @@ func (m *Manifest) dataGatheringJob(namespace string) (*ejv1.ExtendedJob, error)
 			}
 			// Create an init container that copies sources
 			// TODO: destination should also contain release name, to prevent overwrites
-			initContainers = append(initContainers, m.JobSpecCopierContainer(boshJob.Name, releaseName, releaseImage, generateVolumeName("data-gathering")))
+			initContainers = append(initContainers, m.JobSpecCopierContainer(releaseName, releaseImage, generateVolumeName("data-gathering")))
 		}
 
 		// One container per Instance Group
@@ -409,7 +409,7 @@ func (m *Manifest) jobsToInitContainers(igName string, jobs []Job, namespace str
 		if err != nil {
 			return []v1.Container{}, err
 		}
-		initContainers = append(initContainers, m.JobSpecCopierContainer(job.Name, job.Release, releaseImage, "rendering-data"))
+		initContainers = append(initContainers, m.JobSpecCopierContainer(job.Release, releaseImage, "rendering-data"))
 
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/bosh/manifest/kube_converter.go
```diff
@@ -852,3 +819,39 @@ func (m *Manifest) ApplyBPMInfo(kubeConfig *KubeConfig, allResolvedProperties ma
 	}
 	return nil
 }
+
+// AppendToContainerList will append one container to a v1.Container{}
+func (m *Manifest) AppendToContainerList(job Job, releaseImage string, containerReleasePath string, volumeMountName string) []v1.Container {
```
viovanov: This is not appending to anything, you're just returning an array with one `v1.Container`.
The function should be: 
`jobSpecCopierContainer(containerName, releaseImage, volumeMountName, containerMountPath string) v1.Container`

`containerReleasePath` shouldn't be variable - let's always make it `/var/vcap/all-releases/jobs-src`.
Then we need to change the data gatherer and renderer to use that by default.

pkg/bosh/manifest/kube_converter.go
```diff
@@ -889,3 +861,28 @@ func (m *Manifest) ApplyBPMInfo(kubeConfig *KubeConfig, allResolvedProperties ma
 	}
 	return nil
 }
+
+// JobSpecCopierContainer will return a v1.Container{} with the populated field
+// TODO: fix inContainerReleasePath path
```
manno: is that still needed?

pkg/bosh/manifest/kube_converter.go
```diff
@@ -889,3 +861,28 @@ func (m *Manifest) ApplyBPMInfo(kubeConfig *KubeConfig, allResolvedProperties ma
...
+func (m *Manifest) JobSpecCopierContainer(pathTODO string, containerName string, releaseName string, releaseImage string, volumeMountName string, containerMountPath string) v1.Container {
+
+	inContainerReleasePath := filepath.Join(pathTODO, releaseName)
+	// inContainerReleasePath := "/var/vcap/all-releases/jobs-src"
```
manno: Can we remove the commented code, it's probably part of the Git history if we need it again.

qu1queee: sorry, I was testing something, did not expect you to review asap. This will become obsolete with my next change.

pkg/bosh/manifest/kube_converter.go
qu1queee: sorry, I was testing something, did not expect you to review asap. This will become obsolete with my next change.

pkg/bosh/manifest/kube_converter.go
qu1queee: I just addressed your requested changes.

PR #202, opened 2019-04-15
Mention integration test setup in README

manno:


manno:
@viovanov what do you think?

PR #203, opened 2019-04-16
Build binary for core compilation

edwardstudy:
In this scenario, I want to run cf-operator on Darwin but I want to reuse my binary. So I introduced a tool to build cf-operator binary for multiple platforms: https://github.com/mitchellh/gox

And I changed `./bin/up-nobuild` script to use this tool to build binary once and reuse it.

How do you guys think about it?


bin/build-cross-compilation
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
manno: Can we rename this to `build-linux-amd64`?

edwardstudy: @manno Done. Thanks.

bin/up-nobuild
```diff
@@ -8,6 +8,9 @@ set -e
 export DOCKER_IMAGE_TAG=${VERSION_TAG}
 
 if [ -z "$SKIP_IMAGE" ]; then
+  if [ ! -f "binaries/cf-operator-linux-amd64" ]; then
+    bin/build-linux-amd64
```
manno: This script is called 'nobuild', because it doesn't build the binary. So adding code here which does, is confusing :)

Actually this relates to what we talked about in the daily, we have one binary for two very different functions (server and utilities). That's why devs on OSX need a Linux binary in the Dockerimage, but a native one for running the server locally (see below line 20).

Anyhow, if that helps with OSX, we should add it, so devs can use `up-nobuild`. However I'd propose to put the OSX specific behaviour behind another check:

```
if [ `uname -s` = "Darwin" ]; then
  if [ ! -f "binaries/cf-operator-linux-amd64" ]; then
    bin/build-linux-amd64
  fi
fi
```

bin/build-linux-amd64
```diff
@@ -0,0 +1,11 @@
...
+
+set -euo pipefail
+
+GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
+. "${GIT_ROOT}/bin/include/versioning"
+
+BASEDIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+echo "Build binary for image"
+CGO_ENABLED=0 gox -osarch "linux/amd64" -output "${BASEDIR}/binaries/cf-operator-linux-amd64" -mod="vendor" -ldflags="-X code.cloudfoundry.org/cf-operator/version.Version=${ARTIFACT_VERSION}" ./cmd/cf-operator/
```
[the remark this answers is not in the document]
edwardstudy: Gotcha. We didn't need to build multi binaries. I'll change it.

PR #209, opened 2019-04-17
Clean up DataGatherer

manno:
This is just ordering the funcs, replacing an if with a guard statement and cleaning up, it doesn't change the output.

[#165014860](https://www.pivotaltracker.com/story/show/165014860)

### DIFF
--- a/pkg/bosh/manifest/data_gatherer.go
+++ b/pkg/bosh/manifest/data_gatherer.go
@@ -16,7 +16,8 @@ import (
 	"code.cloudfoundry.org/cf-operator/pkg/bosh/bpm"
 )
 
-// DataGatherer gathers data for jobs in the manifest, it handles links and returns a resolved manifest
+// DataGatherer gathers data for jobs in the manifest, it handles links and returns a deployment manifest
+// that only has information pertinent to an instance group.
 type DataGatherer struct {
 	log      *zap.SugaredLogger
 	manifest *Manifest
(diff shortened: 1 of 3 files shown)


pkg/bosh/manifest/data_gatherer.go
```diff
@@ -30,13 +31,17 @@ func NewDataGatherer(log *zap.SugaredLogger, manifest *Manifest) *DataGatherer {
...
+// GenerateManifest will collect different data and return a deployment
+// manifest for an instance group.
+// See docs/rendering_templates.md#calculation-of-required-properties-for-an-instance-group-and-render-bpm
+//
+// Data gathered:
+// * job spec information
+// * job properties
+// * bosh links
+// * bpm yaml file data
+func (dg *DataGatherer) GenerateManifest(baseDir string, namespace string, instanceGroupName string) ([]byte, error) {
```
qu1queee: Not sure if this is fully accurate. While using `manifest` could lead to misinterpretations. AFAIK, there is only one manifest, and that is the `desired manifest`(e.g. after applying ops files).  I liked to have there a reference to the term `datagather`, not sure if something like `GatherAndGenerateData` is more appropriate to the functionality of this cmd.

qu1queee: My comment above should not stop this PR for being merged, just wanted to share some thoughts.

PR #221, opened 2019-04-29
Add extendedstatefulset pvc example

rohitsakala:
[#164684103](https://www.pivotaltracker.com/n/projects/2192232/stories/164684103)


docs/examples/extended-statefulset/README.md
```diff
@@ -19,3 +20,7 @@ When applied on top using `kubectl`, this exemplifies the automatic updating of
 
 This creates 4 `Pods` - 2 in one zone and 2 in another zone.
 
+### exstatefulset_pvcs.yaml
+
+This creates `Statefulset Pods` with `Persistent Volumes Claims` attached to each `Pod`. The presistent volumes claims created, gets re-attached to the new version of `StatefulSet Pods` if the statefulset is updated.
```
viovanov: Typo on `presistent`.
It should also be "the created persistent volume claims get re-attached to the new versions of `StatefulSet` `Pods` when the `ExtendedStatefulSet` is updated.

PR #223, opened 2019-05-02
Docs updates

qu1queee:
Related to env var change, from `CF_OPERATOR_WEBHOOK_HOST` to `CF_OPERATOR_WEBHOOK_SERVICE_HOST`

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ We need to build the binary, which can run out of cluster:
 
 When starting the operator it needs to be reachable from the kubernetes API, so the web hooks work.
 
-    export CF_OPERATOR_WEBHOOK_HOST=<your-public-ip>
+    export CF_OPERATOR_WEBHOOK_SERVICE_HOST=<your-public-ip>
 
 We need to tell the operator which docker image it can use for template rendering:
 
@@ -75,7 +75,7 @@ Go 1.12.2 and install the tool chain:
     export GO111MODULE=on
 
     # set to IP reachable from k8s API
-    export CF_OPERATOR_WEBHOOK_HOST=$(ip -4 a s dev `ip r l 0/0 | cut -f5 -d' '` | grep -oP 'inet \K\S+(?=/)')
+    export CF_OPERATOR_WEBHOOK_SERVICE_HOST=$(ip -4 a s dev `ip r l 0/0 | cut -f5 -d' '` | grep -oP 'inet \K\S+(?=/)')
 
     # optionally, using vendor/ with GO111MODULE=off speeds up docker builds
     go mod vendor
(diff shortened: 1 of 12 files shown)


qu1queee:
@manno I think you are right on the `WEBHOOK_PORT` to `WEBHOOK_SERVICE_PORT`. The reasoning behind the change from `CF_OPERATOR_WEBHOOK_HOST` to `CF_OPERATOR_WEBHOOK_SERVICE_HOST`,  its because when deploying the cf-operator helm chart, we will rely on a [service](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/deploy/helm/cf-operator/templates/service.yaml) so that the `cf-operator` pod can be reached from the api-server. Otherwise jobs, e.g. `data-gathering` will fail as follows:
```
Events:
  Type     Reason        Age   From            Message
  ----     ------        ----  ----            -------
  Warning  FailedCreate  32s   job-controller  Error creating: Internal error occurred: failed calling webhook "mutatepods.example.com": Post https://<SOME-WRONG-IP>:2999/mutate-pods?timeout=30s: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

From the logs above, you will like `SOME-WRONG-IP` to be the right IP, in this case the CLUSTERIP of the service, however you do not know that IP value when deploying the helm chart. 

The trick then, is to rely on environment variables available on the containers(see https://kubernetes.io/docs/concepts/containers/container-environment-variables/#container-environment). Basically, for each service in the namespace, containers will have an adhoc env variable of the form `<SVCNAME_SERVICE_HOST>`, which have the CLUSTERIP as the value, therefore the need of making sure that for the webhook host value in the code, we make use of an environment variable named `CF_OPERATOR_WEBHOOK_SERVICE_HOST`.

manno:
I see, so by switching to 'SERVICE_HOST' we get this automatically inside the cluster. We can get rid of https://github.com/cloudfoundry-incubator/cf-operator/blob/mm/165014860-no-job-index/deploy/helm/cf-operator/templates/operator.yaml#L42-L43 then?

I'd still prefer the same name as the variable, also because we don't 'need' a service. But mapping those variables (somehow) will just add complexity. So I'll merge the SERVICE_* changes.

Can you add it for port, too?

qu1queee:
Ok, let me modify in this PR the changes for the `WEBHOOK_PORT`, then I will request it to be merged.

qu1queee:
This requires https://github.com/cloudfoundry-incubator/cf-operator-ci/pull/43 to be merged.

qu1queee:
@manno i think this one is ready to be merged, can you review?

PR #225, opened 2019-05-03
Move to perl instead of sed for inline replacement

qu1queee:
Inside this script, the sed calls will not work on macOS.
Reason behind is that the `sed -i` requires a value, which needs
to be blank suffix when using the -i on macOS, in order to overwrite
the specified file.
To make this substitution compatible for both linux distros and
macOS, I think we should make use of perl.

### DIFF
--- a/bin/build-helm
+++ b/bin/build-helm
@@ -1,10 +1,10 @@
 #!/bin/bash
 
+set -euo pipefail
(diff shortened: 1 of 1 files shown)


manno:
could have just installed gnu sed :)

PR #226, opened 2019-05-03
Add doc about the BOSHDeployment controller

aduffeck:


docs/controllers/boshdeployment.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Reconcilers
+
+### Deployment Reconciler
+
+Watches for:
+
+- `BOSHDeployment`
+- `ConfigMap`/`Secret` for ops files and the deployment manifest
+- `Pods` created for the BOSHDeployment
```
manno: https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/controllers/boshdeployment/deployment_controller.go#L37-L44

But why?

edwardstudy: Maybe we could clean up this Pod watch if it's OK. I didn't find any Pods belonged to a boshDeployment

docs/controllers/boshdeployment.md
```diff
@@ -0,0 +1,79 @@
...
+> The ["Data Gathering"](https://github.com/cloudfoundry-incubator/cf-operator/tree/master/docs/commands/cf-operator_util_data-gather.md) step generates 2 secrets for each Instance Group:
+>
+> - Instance Group Resolved Properties
+> - Instance Group BPM
+>
+> The "Versioned" `Secret` for Instance Group Resolved Properties is referenced by the Instance Group `ExtendedStatefulSets` and `ExtendedJobs`
+
+Ownership:
+
+TODO document if this will have to keep ownership on something?
```
edwardstudy: Ownership can be used in cleanup phase, so it is needed. AFAIK

docs/controllers/boshdeployment.md
```diff
@@ -0,0 +1,79 @@
+# BOSHDeployment
+
+- [BOSHDeployment](#bosheployment)
```
viovanov: typo in the link

docs/controllers/boshdeployment.md
```diff
@@ -0,0 +1,79 @@
...
+  - [Reconcilers](#reconcilers)
+    - [Deployment Reconciler](#deployment-reconciler)
+    - [Generated Variable Reconciler](#generated-variable-reconciler)
+    - [BPM Reconciler](#bpm-reconciler)
+
+## Description
+
+A BOSH deployment is created from a deployment manifest and optionally ops files.
+
+The deployment manifest is based on a vanilla BOSH deployment manifest,
```
viovanov: Comma at the end of the sentence

docs/controllers/boshdeployment.md
viovanov: There's no reason to watch for pods at this level. If we had to, we'd be doing something wrong.

docs/controllers/boshdeployment.md
viovanov: Based on docs, ownership is used for garbage collection. As a general rule, all resources we _create_ to reconcile a specific object should be owned by that object.

Do we use it for anything else, and if garbage collection is automatic, why do we still need finalizers?

manno: Finalizers are mostly used to remove ownership from objects we don't want to clean up: custom resources created by the user for example.

So when you register ownership on a 'foreign' pod, you also set a finalizer on the new owner. When deleting the owner you can remove finalizer from the owner and the ownership from the pod.

docs/controllers/boshdeployment.md
manno: anybody working on removing it? Or should I create a story?

docs/controllers/boshdeployment.md
edwardstudy: I thought we use ownership to make it happens that when you delete an instance, the instance's dependents(sts, jobs, services, etc) are also deleted. And `finalizers` only controls this process.

1. Without `finalizers`, this deletion is automated
2. Without ownership, this deletion doesn't happen and kube leaves orphans to us

PR #231, opened 2019-05-06
e2e tests for cli

manno:
This adds basic e2e tests for the bosh cli commands.
In the future we could use those, instead of the unit tests.

I think that would be beneficial as data gathering and template rendering also uses Ruby and might write to disk. 

[#165014860](https://www.pivotaltracker.com/story/show/165014860)

### DIFF
--- a/e2e/util_render_template_test.go
+++ b/e2e/util_render_template_test.go
@@ -1,0 +1,53 @@
+package e2e_test
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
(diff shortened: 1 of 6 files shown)


e2e/util_render_template_test.go
```diff
@@ -0,0 +1,53 @@
...
+		session, err = gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
+		return
+	}
+
+	Context("when manifest exists", func() {
+		BeforeEach(func() {
+			manifestPath = "../testing/assets/gatherManifest.yml"
+		})
+
+		It("rendert the instance group template to a file", func() {
```
aduffeck:
```suggestion
		It("renders the instance group template to a file", func() {
```

PR #233, opened 2019-05-07
Add pod UID as part of Job name to avoid re spawning of jobs

rohitsakala:
[#165847530](https://www.pivotaltracker.com/n/projects/2192232/stories/165847530)

### DIFF
--- a/pkg/kube/util/names/names.go
+++ b/pkg/kube/util/names/names.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/kube/util/names/names.go
```diff
@@ -128,21 +129,24 @@ func GetVersionFromVersionedSecretName(name string) (int, error) {
 // JobName returns a unique, short name for a given eJob, pod(if exists) combination
 // k8s allows 63 chars, but the pod will have -\d{6} appended
 // IDEA: maybe use pod.Uid instead of rand
```
manno: Since you implemented this, please remove the comment.

rohitsakala: Removed

PR #234, opened 2019-05-07
Add extended job delete trigger example test case

rohitsakala:
[#165855014](https://www.pivotaltracker.com/n/projects/2192232/stories/165855014)


e2e/kube/examples_test.go
```diff
@@ -66,6 +66,43 @@ var _ = Describe("Examples", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 
+			It("extended-job delete example must work", func() {
```
edwardstudy: @rohitsakala Saw e2e failed at this case

rohitsakala: Solved it.

pkg/kube/controllers/extendedjob/query.go
```diff
@@ -59,3 +65,29 @@ func (q *QueryImpl) MatchState(eJob ejv1.ExtendedJob, podState ejv1.PodState) bo
 	}
 	return false
 }
+
+// IsJobAlreadyExists checks if the job is already created for the pod event match with exjob
+func (q *QueryImpl) IsJobAlreadyExists(ctx context.Context, eJob ejv1.ExtendedJob, podUID types.UID, r *TriggerReconciler) (bool, error) {
```
viovanov: rename to `JobExists` please

viovanov: Also, the comment should be:
```
// JobExists returns true if the Job matching the ExtendedJob has already been created
```

please also rename the `r` parameter to `triggerReconciler`

pkg/kube/util/names/names.go
```diff
@@ -133,7 +133,7 @@ func JobName(eJobName, podName string, podUID types.UID) (string, error) {
 	if podName == "" {
 		suffix = truncate(eJobName, 15)
 	} else {
-		suffix = fmt.Sprintf("%s-%s", truncate(eJobName, 13), truncate(podName, 13))
+		suffix = fmt.Sprintf("%s-%s", truncate(eJobName, 5), truncate(podName, 5))
```
manno: Why does this have to be shorter? Can you add the calculation to the comment, please?
Maybe we should only shorten, if we exceed the 63 char limit, like in `CalculateSecretName`.

I worry those pod names will be even more human-unreadable.

pkg/kube/controllers/extendedjob/trigger_reconciler.go
```diff
@@ -105,16 +105,23 @@ func (r *TriggerReconciler) Reconcile(request reconcile.Request) (result reconci
 
 	for _, eJob := range eJobs.Items {
 		if r.query.MatchState(eJob, podState) && r.query.Match(eJob, *pod) {
-			err := r.createJob(ctx, eJob, podName, pod.GetUID())
+			found, err := r.query.IsJobAlreadyExists(ctx, eJob, pod.GetUID(), r)
```
manno: I might be missing context here, apparently you found a condition where the trigger reconciler tries to start a job again.
To detect this, we now look at all the pods which have been labeled by the eJob. We check if any pod is the same pod as the one the reconciler has been started with. So since the pod has our label, we skip.

I'm wondering why this triggers at all. The predicates should have stopped this? Do you know more about the conditions which trigger that?

Also while skimming over the code, I'm confused by the label, sometimes it's `ejob-name`, then it's `job-name`. Is that intended, or might that be the bug which triggers additional reconciles?

PR #237, opened 2019-05-09
Changing delete event such that duplicate events are removed

rohitsakala:
Multiple delete events were being thrown in the trigger reconciler
code. Ideally, there should be a single delete event. This commit is
defining the delete event as the one, when grace period is set to
zero in pod yaml because that is when the pod is issued KILL signal
and removed from the kube database.

Ref :- https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods


pkg/kube/util/names/names.go
```diff
@@ -128,28 +127,22 @@ func GetVersionFromVersionedSecretName(name string) (int, error) {
 
 // JobName returns a unique, short name for a given eJob, pod(if exists) combination
 // k8s allows 63 chars, but the pod will have -\d{6} appended
-// JobName for which podName and podUID are specified will have the combination of
-// "job" + 9 chars of eJobName + 9 chars of podName + podUID + 5 chars appended by kube
-// JobName for which podName and podUID are not specified will have a combination of
-// "job" + 15 chars of ejobName + randstring + 6 chars appended by kube
-func JobName(eJobName, podName string, podUID types.UID) (string, error) {
+// IDEA: maybe use pod.Uid instead of rand
```
edwardstudy: Why we reverted combination of Uid

manno: Apparently we don't need it, Rohit found a way to improve the predicates and the pod state matcher. So we don't get triggered multiple times anymore.

It's not a good idea to do a string match on the name, if we want the triggering pod UID, we should put it in a label, like this: https://github.com/cloudfoundry-incubator/cf-operator/pull/new/mm/triggering-pod-label

PR #238, opened 2019-05-10
Add `triggering-pod` label on job pods

manno:
Feature was documented, but not implemented. Also explain where 'job-name' label comes from.

### DIFF
--- a/pkg/kube/controllers/extendedjob/trigger_reconciler.go
+++ b/pkg/kube/controllers/extendedjob/trigger_reconciler.go
@@ -104,7 +104,7 @@ func (r *TriggerReconciler) Reconcile(request reconcile.Request) (result reconci
 
 	for _, eJob := range eJobs.Items {
 		if r.query.MatchState(eJob, podState) && r.query.Match(eJob, *pod) {
-			err := r.createJob(ctx, eJob, podName)
+			err := r.createJob(ctx, eJob, podName, string(pod.UID))
 			if err != nil {
 				if apierrors.IsAlreadyExists(err) {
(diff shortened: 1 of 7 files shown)


pkg/kube/controllers/extendedjob/trigger_reconciler.go
```diff
@@ -119,13 +119,14 @@ func (r *TriggerReconciler) Reconcile(request reconcile.Request) (result reconci
...
 
-func (r *TriggerReconciler) createJob(ctx context.Context, eJob ejv1.ExtendedJob, podName string) error {
+func (r *TriggerReconciler) createJob(ctx context.Context, eJob ejv1.ExtendedJob, podName string, podUID string) error {
 	template := eJob.Spec.Template.DeepCopy()
 
 	if template.Labels == nil {
 		template.Labels = map[string]string{}
 	}
 	template.Labels["ejob-name"] = eJob.Name
+	template.Labels["triggering-pod"] = podUID
```
edwardstudy: could we constantize label keys, such as: https://github.com/cloudfoundry-incubator/cf-operator/blob/d2c424b3ecd3acce4efdfe2b252dc9cb61cb11cd/pkg/kube/apis/extendedjob/v1alpha1/types.go#L17-L20

manno: Maybe it's a little bit beyond the scope of this PR to replace all the keys, but since we have to test it all again anyhow....

aduffeck:
Tests passed locally.

PR #240, opened 2019-05-13
Extract data gathering and var interpolation jobs into a factory

aduffeck:


manno:
Tested locally and it passes all the tests.

PR #242, opened 2019-05-13
Improvements in the helm chart definition

qu1queee:
Make usage of the `values.yaml` in different chart templates
Do not use latest for the cf-operator img
Fix script, that overrides the chart.

### DIFF
--- a/bin/build-helm
+++ b/bin/build-helm
@@ -12,7 +12,8 @@ filename="${output_dir}/${ARTIFACT_NAME}-${ARTIFACT_VERSION}.tgz"
 [ -d "${output_dir}" ] && rm -r "${output_dir}"
 cp -r "${GIT_ROOT}/deploy/helm" "${output_dir}"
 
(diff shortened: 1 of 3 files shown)


bin/build-helm
```diff
@@ -12,7 +12,8 @@ filename="${output_dir}/${ARTIFACT_NAME}-${ARTIFACT_VERSION}.tgz"
 [ -d "${output_dir}" ] && rm -r "${output_dir}"
 cp -r "${GIT_ROOT}/deploy/helm" "${output_dir}"
 
-perl -pi -e "s|repository: .*|repository: ${OPERATOR_DOCKER_ORGANIZATION}/cf-operator|g" "${output_dir}/cf-operator/values.yaml"
+perl -pi -e "s|repository: .*|repository: cf-operator|g" "${output_dir}/cf-operator/values.yaml"
```
manno: Those substitutions probably don't have to be greedy :)

PR #249, opened 2019-05-15
E2E tests for helm/kube, this includes:

qu1queee:
- Modify kube/e2e tests, to deploy the cf-operator helm charts
  directly in a kube cluster via helm.  This will first setup a suitable env, where helm/tiller should 
  be both running. Also, make sure the generated resources are deleted
  at the end of the suite.
- Add new Makefile target, `test-helm-e2e` to run the above mentioned
  e2e tests.
- Add `test-kube-e2e` script, so that the new `test-helm-e2e` target can
  run it.
- Add readinessProbe to the `cf-operator` chart template, so that when installing, helm can wait till the cmd of the readinessProbe its executed. This ensures, that none e2e test will run prior to the cf-operator pod being in `RUNNING` state.
- Ensure the `cf-operator` chart template, populates the `imagePullPolicy` key via the values.yaml.

This runs everything in "default" ns. 

There is a dependency with https://github.com/cloudfoundry-incubator/cf-operator-ci/pull/50, it needs to be merged first, in order for this PR to work properly.


bin/test-kube-e2e
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -eu
+
+GIT_ROOT="${GIT_ROOT:-$(git rev-parse --show-toplevel)}"
+. "${GIT_ROOT}/bin/include/versioning"
+
+export DOCKER_IMAGE_TAG=${VERSION_TAG}
+export TEST_NAMESPACE="default"
```
edwardstudy: Could we use "test$(date +%s)" namespace to do e2e as integration-test does? If tests are triggered in parallel, we should keep tests didn't influence each other.

qu1queee: Hi, this is intended to run on another cluster as part of the CI. If I run this locally, I usually do all tests sequentially. Do you have a use case where you run locally everything in parallel? 
I think I did not test this running in another namespace besides default.

qu1queee: @edwardstudy if you consider that this test should use a custom ns, please request it as a change, so I can include it in this PR :)

edwardstudy: I understood that e2e tests will be run in CI and we need to trigger CI builds in parallel.

manno: @qu1queee I think we should support a test namespace and helm installing the operator into it. Not sure if that's in the scope of your story? Anyhow let's do that in a separate PR

e2e/kube/suite_test.go
```diff
@@ -16,28 +21,134 @@ func TestE2E(t *testing.T) {
...
+	}
+
+	return nil
+}
+
+// SetUpEnvironment ensures helm binary can run,
+// being able to reach tiller, and eventually it
+// will install the cf-operator chart.
+func SetUpEnvironment() error {
+
```
manno: Why is everybody doing newlines after the func?

e2e/kube/suite_test.go
```diff
@@ -16,28 +21,134 @@ func TestE2E(t *testing.T) {
...
+		}
+	}
+
+	dir, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Install the helm chart in default ns
+	_, err = RunBinary(helmCmd, "install", fmt.Sprintf("%s%s", dir, "/../../helm/cf-operator"),
```
manno: I thought we need to run `make build-helm` first and install the tgz?

qu1queee: it happens in the bin/test-kube-e2e

bin/test-kube-e2e
qu1queee: its not, but I agree we should. I prefer to have it in another PR.

qu1queee:
Please do not merge this PR yet.

bin/test-kube-e2e
qu1queee: @edwardstudy you were right. I was not aware we run multiple PRs check in parallel. In any case I already add the test ns support.

qu1queee:
This PR will generate a race condition between pipelines, e.g. (cf-operator-check vs cf-operator), when both runs the e2e tests. The issue is that we do a `helm install` in the test, and in order for the helm release installation to work, the CRDs  should not be in place.(e.g. to check this `kubectl get crds`). Otherwise, helm will fail to install. The only solution I see, is to introduce a locking mechanism in our CI, so that only one pipeline runs the e2e at a time. I did a prototype of this based on the [official concourse release](https://github.com/concourse/pool-resource) , and this seems to work well.
We will need an external repo, where we can store the metadata of the lock owner, e.g. https://github.com/qu1queee/lock-it . If you are interested on checking this test, you can try it with the following 2 pipelines(where only one should run the `build` task at a time):
- https://ci.flintstone.cf.cloud.ibm.com/teams/containerization/pipelines/cf-operator-lock-a
- https://ci.flintstone.cf.cloud.ibm.com/teams/containerization/pipelines/cf-operator-lock-b

qu1queee:
I added some more commits, in order to leverage the helm charts definition, for the `clusteRole`, `ClusterRoleBinding` and `ServiceAccount` chart templates. 

These are:
- As stated by helm best practices, avoid the usage of plain values, but better to use references to the values.yml or default values. This will allows to generate an specific set of RBAC´s + serviceAccount based on the release name of the chart.
- Ensure the ServiceAccount is binded to the namespace where the chart is deploy.

The above key points,  will allow us to deploy different charts into the same cluster at the same time. We do not need anymore any sort of locking mechanism.

PR #266, opened 2019-05-22
Define a readinessProbe endpoint for the cf-operator

qu1queee:
- Add new probe pkg to give back 200 or 500 http codes depending
  on the current state of the operator. e.g. if the manager is
  already started, which means that the webhook is already created.

- Modify the cf-operator template chart, to make use of the new
  /readyz route.


cmd/internal/root.go
```diff
@@ -21,6 +24,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
+const (
+	readinessProbeHostPort = "0.0.0.0:8080"
```
viovanov: We should do this on the same server as the webhook.

viovanov:
We have to try and use the same server to serve the health check. 
Otherwise, we'll need a new tunnel for CI.
And after all, we want to check the actual server.

PR #268, opened 2019-05-22
Enabled builds with cf-operator as submodule

f0rmiga:
Part of [#166081131](https://www.pivotaltracker.com/story/show/166081131)

The git commands can get messy when cf-operator is used as a git submodule. This PR enables overriding ARTIFACT_VERSION, especially for development purposes.


f0rmiga:
Closing as we decided to take another path on this.

PR #270, opened 2019-05-23
Reserve preset bpm info in desired manifest

edwardstudy:
[#166115594](https://www.pivotaltracker.com/story/show/166115594)

### DIFF
--- a/pkg/bosh/manifest/data_gatherer_test.go
+++ b/pkg/bosh/manifest/data_gatherer_test.go
@@ -95,21 +95,37 @@ var _ = Describe("DataGatherer", func() {
 				Expect(bpm.Processes[0].Env["FOOBARWITHSPECDEPLOYMENT"]).To(Equal("cf"))
 			})
 
-
-
-			Context("when manifest presets bpm info", func() {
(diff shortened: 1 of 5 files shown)


pkg/bosh/manifest/data_gatherer_test.go
```diff
@@ -94,6 +94,24 @@ var _ = Describe("DataGatherer", func() {
 				Expect(bpm.Processes[0].Env["FOOBARWITHSPECADDRESS"]).To(Equal("cf-log-api-0.default.svc.cluster.local"))
 				Expect(bpm.Processes[0].Env["FOOBARWITHSPECDEPLOYMENT"]).To(Equal("cf"))
 			})
+
```
aduffeck: Could you remove the superfluous newlines?

aduffeck:
I wonder if we should allow for overriding only one of the processes in a multi-process BPM scenario. The story is not very specific in that regard, but I think it's not unlikely that you'd want to only override one of the processes.

edwardstudy:
@aduffeck This's a good pointer. I thought it can be done and minimal object is one process in list.

PR #278, opened 2019-05-26
Removed Kubernetes module

f0rmiga:
Removed Kubernetes as a Go Module since we were using only 2 functions.
The Kubernetes repository is not supposed to be depended on, we should rather depend on the individual packages that the github.com/kubernetes org provides.


manno:
Great to get rid of that big dependency at the cost of  copying just one function :)

> The Kubernetes repository is not supposed to be depended on

Out of curiosity, is that documented upstream somewhere?

PR #290, opened 2019-05-30
Apply user specified labels and annotations on created resources

edwardstudy:
https://www.pivotaltracker.com/story/show/165909434


pkg/bosh/manifest/kube_resources.go
```diff
@@ -54,6 +53,24 @@ func (kc *KubeConverter) BPMResources(manifestName string, version string, insta
 	res := &BPMResources{}
 
 	for _, ig := range instanceGroups {
+		// Override labels and annotation with operator-owned metadata
```
viovanov: can be simplified

```
		// Override labels and annotation with operator-owned metadata
		if ig.Env.AgentEnvBoshConfig.Agent.Settings.Labels == nil {
			ig.Env.AgentEnvBoshConfig.Agent.Settings.Labels = map[string]string{}
		}
		
		ig.Env.AgentEnvBoshConfig.Agent.Settings.Labels[LabelDeploymentName] = manifestName
		ig.Env.AgentEnvBoshConfig.Agent.Settings.Labels[LabelInstanceGroupName] = ig.Name

```

pkg/bosh/manifest/instance_group.go
```diff
@@ -96,10 +96,16 @@ type JobDir struct {
 	TmpfsSize string `yaml:"tmpfs_size,omitempty"`
 }
 
+// Agent from BOSH deployment manifest
```
manno: How about also explaining what these settings are for, something like:
```suggestion
// AgentSettings from BOSH deployment manifest. These annotations and labels are added to ....
```

PR #303, opened 2019-06-04
Refactor and magic volume strings removal

f0rmiga:
- Pulled generateJobContainers out JobsToContainers.
- Removed magic strings for volumes.


pkg/bosh/manifest/consts.go
```diff
@@ -0,0 +1,26 @@
+package manifest
```
manno: Can we name this file `volume.go` or something more descriptive? We do have other constants and they are not in this file. Also files named like 'models', 'constants', 'interfaces' have the tendency to grow into collections of things, which are not really related conceptually.

Maybe we can have a file for all the volume related structs. There are also two concepts in that package, which haven't been modeled explicitly: the 'input BOSH' manifest and the resulting structs after kube conversion.

I think in the beginning we even wanted to avoid importing k8s at all into the bosh packages and now we have the factories, resolvers  and converters all in there. (@viovanov ?)

pkg/bosh/manifest/container_factory.go
```diff
@@ -156,22 +86,92 @@ func (c *ContainerFactory) JobsToContainers(jobs []Job) ([]corev1.Container, err
 	return containers, nil
 }
 
+func (c *ContainerFactory) generateJobContainers(job Job, jobImage string) ([]corev1.Container, error) {
+	boshJobName := job.Name
+	containers := []corev1.Container{}
+	template := corev1.Container{
```
manno: Uhm, while you're at it, could you extract that into a func? I think all the other containers have one (`templateRenderingContainer`, etc...)

pkg/bosh/manifest/consts.go
viovanov: yes, the desire was to only work with structures, and not perform any Kube API action from this package

viovanov:
yay for constants

pkg/bosh/manifest/consts.go
manno: well, I guess we are still fine, as we don't use the API. :+1:

f0rmiga: _busted_

You are absolutely right. Whereas it is relatively common to find `const*.go` files in well-known Go repositories (e.g. https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/types/constants.go), you put it really well, remembering me of Alan Donovan's book on how to name things.

f0rmiga: For the structs part, we only instantiate upstream structs, so no actions in this regard to be taken.

PR #305, opened 2019-06-05
Add Bazel integration

f0rmiga:
## Description

Added Bazel build system integration for building the binary, building the Docker image in a deterministic way, publishing the image to the remote Docker registry and unit testing.

Left the old build system untouched.

The first time Bazel runs, it will download the dependencies and cache them in a very optimized way, making the next operations blazing fast. The invalidation of the cache happens on the change of the dependencies, causing a build on the reverse dependency tree. I.e. a change to a .go file will trigger the build of the targets that depend on the target that uses that file.

## Test plan

- The current build flow **SHOULD NOT** be affected.
- Check that all the Go targets are up-to-date with Gazelle: `bazel run //:gazelle`. No changes to the `BUILD.bazel` files should happen.
- Run the unit tests: `bazel test -- //... -//integration/... -//e2e/...`.
- Run the cf-operator on Minikube: `skaffold run --namespace scf`. Skaffold will trigger Bazel and then use the Helm charts to deploy. To validate the optimized cache, re-create the Minikube instance and perform another `skaffold run`. The cf-operator image should not be built again, only loaded into the Minikube Docker daemon (performed automatically by Skaffold).
- Build the image using `USE_BAZEL=1 bin/build-image`.


Makefile
```diff
@@ -3,64 +3,64 @@
 all: tools build test
 
 up:
-	bin/up
+	bin/up.sh
```
manno: I'd rather not change that. Upstream projects also omit the extension: https://github.com/cloudfoundry/bosh-cli/tree/master/bin
It's also common for scripts in bin folders on Linux. Any reason to change it?

f0rmiga: I see. Bazel got confused with `bin/build`, thinking that it is a BUILD file. I'll see what I can do to prevent that other than changing the name of the files.

bin/bazel_status
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
f0rmiga: This file is the bridge between Bazel and the current setup for stamping the binary and the image org/tag. It is used by Bazel with `--workspace_status_command=bin/bazel_status`. Ref https://docs.bazel.build/versions/master/user-manual.html#flag--workspace_status_command.

cmd/cf-operator/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+
+go_binary(
+    name = "cf-operator",
+    embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+)
+
+go_image(
+    name = "image",
+    base = "@cf_operator_base//image",
```
f0rmiga: The base cf-operator image is referenced here and only the binary is appended as a layer on top of the image, optimizing the cache.

cmd/cf-operator/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+    embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+    goos = "linux",
+    goarch = "amd64",
+    pure = "on",
+)
+
+container_push(
+   name = "push_image_dockerhub",
```
f0rmiga: With `bazel run //cmd/cf-operator:push_image_dockerhub`, the image can be published to the registry.

deploy/helm/cf-operator/templates/operator.yaml
```diff
@@ -19,8 +19,6 @@ spec:
           ports:
           - containerPort: 60000
             name: metrics
-          command:
-          - cf-operator
```
f0rmiga: We trust in the entrypoint set by the Dockerfile of by Bazel.

WORKSPACE
```diff
@@ -0,0 +1,637 @@
...
+
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_pull",
+)
+
+container_pull(
+    name = "cf_operator_base",
+    registry = "index.docker.io",
+    repository = "thulioassis/cf-operator-base",
```
f0rmiga: This should be replaced by the `cfcontainerization` org if this PR gets approved.

f0rmiga:
Closing this for now. We will revisit in the future.

PR #309, opened 2019-06-05
[#166387510] BOSH pre-start as init containers

f0rmiga:
[#166387510](https://www.pivotaltracker.com/story/show/166387510)

### DIFF
--- a/pkg/bosh/manifest/container_factory.go
+++ b/pkg/bosh/manifest/container_factory.go
@@ -31,23 +31,59 @@ func NewContainerFactory(manifestName string, igName string, releaseImageProvide
 }
 
 // JobsToInitContainers creates a list of Containers for corev1.PodSpec InitContainers field
-func (c *ContainerFactory) JobsToInitContainers(jobs []Job) ([]corev1.Container, error) {
-	initContainers := []corev1.Container{}
+func (c *ContainerFactory) JobsToInitContainers(jobs []Job, hasPersistentDisk bool) ([]corev1.Container, error) {
+	copyingSpecsInitContainers := make([]corev1.Container, 0)
+	boshPreStartInitContainers := make([]corev1.Container, 0)
+	bpmPreStartInitContainers := make([]corev1.Container, 0)
 
-	// one init container for each release, for copying specs
-	doneReleases := map[string]bool{}
+	copyingSpecsUniq := map[string]struct{}{}
 	for _, job := range jobs {
-		if _, ok := doneReleases[job.Release]; ok {
-			continue
-		}
-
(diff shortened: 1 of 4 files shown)


pkg/bosh/manifest/container_factory.go
```diff
@@ -31,23 +31,39 @@ func NewContainerFactory(manifestName string, igName string, releaseImageProvide
...
+			copyingSpecsInitContainers = append(copyingSpecsInitContainers, copyingSpecsInitContainer)
+		}
+
+		jobPreStart := filepath.Join(VolumeJobsDirMountPath, job.Name, "bin", "pre-start")
+		boshPreStartInitContainer := corev1.Container{
+			Name:         fmt.Sprintf("bosh-pre-start-%s", job.Name),
+			Image:        jobImage,
+			VolumeMounts: instanceGroupVolumeMounts(),
+			Command:      []string{"/bin/sh", "-c"},
+			Args:         []string{fmt.Sprintf(`test -x "%[1]s" && "%[1]s"`, jobPreStart)},
```
manno: If the pre-start hook is not available (missing execute permission, file not found) we won't fail. 

@viovanov  However we might want to fail if the script is missing.

viovanov: yes, we want to fail

manno: Now I understand. The pre-start is always rendered to that file (`bin/pre-start`), so if it's missing there is no pre-start configured.
I think then it's fine to rely on our rendering to do the right thing?

Also, while working on the integration test for BPM pre-start scripts, I found the nats-release, for which we already have a manifest (https://github.com/cloudfoundry-incubator/cf-operator/blob/master/testing/boshmanifest/manifests.go#L94). 
While working on the integration test for BPM pre-start scripts, I found the nats-release, for which we already have a manifest (https://github.com/cloudfoundry-incubator/cf-operator/blob/master/testing/boshmanifest/manifests.go#L94). 
It only has a BOSH a pre-start script, so I can't use it, but it should work here.

f0rmiga: According to the documentation, the script **can** exist, meaning it is not required. I should change to:
```
/bin/sh -c 'if [ -x "%[1]s" ]; then "%[1]s"; fi'
```
If the script exists and is executable, try executing it, otherwise, return 0. I was tempted to add `set -o errexit`, but I think it is up to the script to set it.

f0rmiga: E.g. https://github.com/cloudfoundry/nats-release/blob/master/jobs/nats/templates/pre-start.erb handles `errexit`.

PR #312, opened 2019-06-06
Minikube config file for local dev flow

qu1queee:
Adding  a havener configuration file that packs a complete development flow, for testing changes in the cf-operator.  Details are described in the *.yaml itself, but mainly it:
- compiles a binary
- builds a Docker img using the above binary
- Installs a helm chart using the above docker img

This file currently relies on minikube as the kubernetes cluster.

We do not have a story for this, only from retros(Action items). This setup is helpful for me, when I´m doing changes in the cf-operator code, and I want to see the behavior in a kube cluster. Not saying is the way to go.

### DIFF
--- a/dev-env-havener.yaml
+++ b/dev-env-havener.yaml
@@ -1,0 +1,78 @@
+#
+#
+# The dev-env-havener.yaml packs a development flow 
+# for a running minikube into a single configuration file. 
+# This flow involves:
+# - compiling a binary
+# - building a Docker img using the above binary
+# - Installing a helm chart using the above docker img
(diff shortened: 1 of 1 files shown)


cf-operator-deployment.yaml
```diff
@@ -0,0 +1,78 @@
+#
```
viovanov: The name of the file is too generic - people could think it's something else.
How about `dev-env-havener.yaml` ?

qu1queee: sounds good.

PR #316, opened 2019-06-10
Support for BPM volumes

qu1queee:
[#166404375](https://www.pivotaltracker.com/story/show/166404375)

This includes:
- bpm ephemeral disk support (uses a emptyDir volume)
- bpm persistent disk support (ignored for the moment)
- bpm additional volumes support (uses a emptyDir volume)
- bpm unrestricted volumes support (uses a emptyDir volume)

For both `additional_volumes` and ` unrestricted_volumes` , paths under `/var/vcap/store/` are not
supported, until we find a better way to handle persistentVolumesClaim.

This PR generates the volumeMounts for the above types, for the
containers(corev1.PodSpec)  and initContainers(only bpm hooks), and also a matching Volume for the pod definition.

### DIFF
--- a/docs/from_bosh_to_kube.md
+++ b/docs/from_bosh_to_kube.md
@@ -340,16 +340,17 @@ The following subsections describe the mapping of BPM configuration into contain
 
 ### Resources
 
-| Bosh                   | Kube Pod Container                                             |
-| ---------------------- | -------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


viovanov:
looks good - please also [update docs](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/from_bosh_to_kube.md#bpm)

qu1queee:
@viovanov done, docs updated

PR #317, opened 2019-06-11
Convert container names

manno:
[#166502851](https://www.pivotaltracker.com/story/show/166502851)

### DIFF
--- a/pkg/bosh/manifest/container_factory.go
+++ b/pkg/bosh/manifest/container_factory.go
@@ -61,11 +61,15 @@ func (c *ContainerFactory) JobsToInitContainers(jobs []Job, hasPersistentDisk bo
 		}
 		for _, process := range bpmConfig.Processes {
 			if process.Hooks.PreStart != "" {
+				container := bpmPrestartInitContainer(process.Name, jobImage, process.Hooks.PreStart)
+
 				bpmVolumes, err := generateBPMVolumes(process, job.Name)
 				if err != nil {
 					return []corev1.Container{}, err
 				}
-				bpmPreStartInitContainers = append(bpmPreStartInitContainers, bpmPrestartInitContainer(process.Name, jobImage, process.Hooks.PreStart, bpmVolumes))
+				container.VolumeMounts = append(instanceGroupVolumeMounts(), bpmVolumes...)
+
+				bpmPreStartInitContainers = append(bpmPreStartInitContainers, container)
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/bosh/manifest/container_factory.go
```diff
@@ -51,22 +51,7 @@ func (c *ContainerFactory) JobsToInitContainers(jobs []Job, hasPersistentDisk bo
...
-		}
-		if hasPersistentDisk {
-			persistentDisk := corev1.VolumeMount{
-				Name:      VolumeStoreDirName,
-				MountPath: VolumeStoreDirMountPath,
-			}
-			boshPreStartInitContainer.VolumeMounts = append(boshPreStartInitContainer.VolumeMounts, persistentDisk)
-		}
-		boshPreStartInitContainers = append(boshPreStartInitContainers, boshPreStartInitContainer)
+		boshPreStartInitContainers = append(boshPreStartInitContainers, boshPreStartInitContainer(job.Name, jobImage, hasPersistentDisk))
```
f0rmiga: I'm not comfortable with all these memory copying. We are passing structs all around by value. On one hand, it prevents garbage collection from happening, on the other hand, it can get wild with structs growing. IMO we either leave the way it was or we start passing the pointers around.

pkg/kube/util/names/names.go
```diff
@@ -82,6 +74,30 @@ func CalculateEJobOutputSecretPrefixAndName(secretType DeploymentSecretType, dep
 	return prefix + ".", finalName
 }
 
+var allowedKubeChars = regexp.MustCompile("[^-a-z0-9]*")
+
+// ContainerName k8s valid container name: [a-z0-9]([-a-z0-9]*[a-z0-9])?
+func ContainerName(name string) string {
```
f0rmiga: How about calling it `Sanitize`?

manno: It doesn't seem to be in the core. Do you mean this one for HTML https://github.com/kennygrant/sanitize? 
I wouldn't bring another dependency in when a simple regex is enough.

f0rmiga: Oh no, just the naming. `Sanitize` sounds more correct than `ContainerName`.

pkg/bosh/manifest/container_factory.go
manno: Yes the function call returns a struct, but I don't see a need to optimize this. The number of elements is rather small and it's not that costly.
 
More importantly, I think it's totally worth passing the struct for readability, since pointers signal a mutable relationship to the reader.  We also avoid dealing with nil.

https://medium.com/@vCabbage/go-are-pointers-a-performance-optimization-a95840d3ef85

If you don't put the struct into the HEAP, there is no need for garbage collection. I wouldn't use pointers for performance reasons (well, unless it's performance critical and I benchmarked first).

If you think it's really hurting us, please provide a benchmark and add a separate story to optimize the container factory.

pkg/kube/util/names/names.go
manno: Hm, yes I can rename it. Sanitize sounds great.

 I was just wondering if it's too early to find a general name, the func is only used for the container names now. As far as I understand the other resources would also benefit from this, but are implemented slightly different now, i.e. services have a dash in between name and md5. So I expected some clean up in the future...

f0rmiga: Yes, we can reuse it for various k8s resource naming. Thanks for renaming it.

pkg/bosh/manifest/container_factory.go
f0rmiga: I don't think it's really hurting us at this point. It could though. I'll leave it for the future.

PR #321, opened 2019-06-12
remove ownership/finalizers from ests

edwardstudy:
[#166266255](https://www.pivotaltracker.com/story/show/166266255)


edwardstudy:
Need CI to do PVC test.

PR #323, opened 2019-06-13
Add target to print the current tag of the cf-operator image

viovanov:


bin/print-tag
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+GIT_ROOT="${GIT_ROOT:-$(git rev-parse --show-toplevel)}"
+. "${GIT_ROOT}/bin/include/versioning"
+. "${GIT_ROOT}/.envrc"
+
+if [ -z ${DOCKER_IMAGE_TAG+x} ]; then
```
f0rmiga: Should this entirely be part of `versioning`? I see the same logic here https://github.com/cloudfoundry-incubator/cf-operator/blob/master/bin/build-image.

viovanov:
one can use `DOCKER_IMAGE_TAG` to always use the same thing in dev

PR #324, opened 2019-06-13
Revert "Added chown 1000:1000 cmd for the /var/vcap/..."

viovanov:
Reverts cloudfoundry-incubator/cf-operator#322

### DIFF
--- a/docs/from_bosh_to_kube.md
+++ b/docs/from_bosh_to_kube.md
@@ -34,21 +34,17 @@
 
 ## Missing Features
 
-1. Canary support in ExtendedStatefulSets.
+1. Canary support in ExtendedStatefulSets
 
 ## High-level Direction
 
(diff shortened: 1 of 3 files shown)


viovanov:
@f0rmiga 

We're seeing this

```
$ k logs nats-deployment-nats-v1-0 -c create-dirs-nats
chown: changing ownership of '/var/vcap/data/nats': Operation not permitted
chown: changing ownership of '/var/vcap/data/sys/log/nats': Operation not permitted
chown: changing ownership of '/var/vcap/data/sys/run/nats': Operation not permitted
chown: changing ownership of '/var/vcap/sys/log/nats': Operation not permitted
chown: changing ownership of '/var/vcap/sys/run/nats': Operation not permitted
```

Also, markdown links seem to be broken.

PR #330, opened 2019-06-14
EJob versioned secrets triggers

manno:
Adds triggers for versioned secrets. 

Updates work, but now it triggers too often, so we miss `v2` and have `v3` instead...

https://drive.google.com/file/d/13op-0R9h3ZLmjFWDwC4DTXPwn7jjEG1-/view?usp=sharing

[#166541753](https://www.pivotaltracker.com/story/show/166541753)

### DIFF
--- a/pkg/kube/util/versionedsecretstore/versioned_secret.go
+++ b/pkg/kube/util/versionedsecretstore/versioned_secret.go
@@ -1,0 +1,57 @@
+package versionedsecretstore
+
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// IsVersionedSecret returns true if the secret has a label identifying it as versioned secret
+func IsVersionedSecret(secret corev1.Secret) bool {
+	secretLabels := secret.GetLabels()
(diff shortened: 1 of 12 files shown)


edwardstudy:
Sure. I think it can repair flaky CI:
https://ci.flintstone.cf.cloud.ibm.com/teams/containerization/pipelines/cf-operator-check/jobs/test/builds/858?#L5d0337c8:420:427

pkg/kube/util/versionedsecretstore/versioned_secret.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if kind, ok := secretLabels[LabelSecretKind]; ok && kind == VersionSecretKind {
+		return true
+	}
+
+	return false
+}
+
+// UnversionedName returns the unversioned name of a secret by removing the /-v\d+/ suffix
+func UnversionedName(name string) (string, error) {
```
edwardstudy: We can reuse this function: https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/util/names/names.go#L121

PR #331, opened 2019-06-14
Refactored JobSpec Consumes and Provides

f0rmiga:
Also fixed error message for missing provider.


pkg/bosh/manifest/data_gatherer.go
```diff
@@ -385,7 +385,7 @@ func generateJobConsumersData(currentJob *Job, jobReleaseSpecs map[string]map[st
 
 		link, hasLink := jobProviderLinks.Lookup(consumes.Type, consumesName)
 		if !hasLink && !consumes.Optional {
-			return fmt.Errorf("cannot resolve non-optional link for consumer %s", consumesName)
+			return fmt.Errorf("cannot resolve non-optional link for provider %s", consumesName)
```
edwardstudy: Why don't we change `consumesName` to `providerName` also?

f0rmiga: I did a broader cleanup on providers/links.

PR #349, opened 2019-06-19
Skip creating PersistentVolumeClaim if exists

edwardstudy:
[#166555413](https://www.pivotaltracker.com/story/show/166555413)


edwardstudy:
CI issue could not be reproduced locally

PR #354, opened 2019-06-21
Fix integration test - remove unwanted asserts

rohitsakala:


integration/environment/machine.go
```diff
@@ -722,6 +722,8 @@ func (m *Machine) WaitForJobsDeleted(namespace string, labels string) error {
 func (m *Machine) JobExists(namespace string, name string) (bool, error) {
 	_, err := m.Clientset.BatchV1().Jobs(namespace).Get(name, metav1.GetOptions{})
 	if err != nil {
+
+		fmt.Println(err)
```
edwardstudy: Should we print this error?

rohitsakala: Removed.

PR #356, opened 2019-06-24
Generate files for initContainers when dumping logs

qu1queee:
[#166661635](https://www.pivotaltracker.com/story/show/166661635)

Generate files for initContainers when dumping logs
- For every pod, add an initcontainer dir, with a kube.log file

Modify initContainer cmd and args, to have a common struct
- Use Command and Args fields, when defining the corev1.Container
- Use '-x'  to dump an stack of the executed cmd into the pod logs

### DIFF
--- a/pkg/bosh/manifest/container_factory.go
+++ b/pkg/bosh/manifest/container_factory.go
@@ -212,9 +212,11 @@ func jobSpecCopierContainer(releaseName string, jobImage string, volumeMountName
 			},
 		},
 		Command: []string{
-			"bash",
-			"-c",
-			fmt.Sprintf(`mkdir -p "%s" && cp -ar %s/* "%s"`, inContainerReleasePath, VolumeJobsSrcDirMountPath, inContainerReleasePath),
+			"/bin/sh",
+		},
+		Args: []string{
+			"-xc",
+			fmt.Sprintf("mkdir -p %s && cp -ar %s/* %s", inContainerReleasePath, VolumeJobsSrcDirMountPath, inContainerReleasePath),
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


pkg/bosh/manifest/container_factory.go
```diff
@@ -212,9 +213,11 @@ func jobSpecCopierContainer(releaseName string, jobImage string, volumeMountName
...
 		},
 		Command: []string{
-			"bash",
-			"-c",
-			fmt.Sprintf(`mkdir -p "%s" && cp -ar %s/* "%s"`, inContainerReleasePath, VolumeJobsSrcDirMountPath, inContainerReleasePath),
+			"/bin/sh",
+		},
+		Args: []string{
+			"-xc",
+			specCopierCmd,
```
viovanov: Why a new variable?

pkg/bosh/manifest/container_factory.go
```diff
@@ -276,9 +287,14 @@ func createDirContainer(name string, jobs []Job) corev1.Container {
...
-		Env:     []corev1.EnvVar{},
-		Command: []string{"/bin/sh", "-c"},
-		Args:    []string{fmt.Sprintf("mkdir -p %s", strings.Join(dirs, " "))},
+		Env: []corev1.EnvVar{},
+		Command: []string{
+			"/bin/sh",
+		},
+		Args: []string{
+			"-xc",
+			createDirCmd,
```
viovanov: again, it makes it hard to follow the code

pkg/bosh/manifest/container_factory.go
qu1queee: when I changed this, I was thinking on improving test cases, by having a fixed command constant that can be used in the code and test cases. I think this cannot be a constant and I forgot to rollback this change, will do it now.

PR #364, opened 2019-06-30
Support Kubernetes Affinity

edwardstudy:
[#165909121](https://www.pivotaltracker.com/story/show/165909121)

This PR is working on:
- Converted manifest YAML to JSON and then unmarshaled to struct because kube API objects work with JSON primarily such as `Affinity`
- Added `json` tags to Manifest struct and supported JSON `inline` tags for `JobProperties`

After that, we can support `Affinity` concisely:
```
instance_groups:
- name: bpm
  instances: 1
  jobs:
  - name: test-server
    release: bpm
    properties: ...
  env:
    bosh:
      agent:
        settings:
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                - matchExpressions:
                  - key: kubernetes.io/unit-test-az-name
                    operator: In
                    values:
                    - unit-test-az1
                    - unit-test-az2
```

**Notice**: If you want to decode/encode Manifest struct and substruct, you should use package `"github.com/ghodss/yaml"` as kubectl does.

References:
- https://github.com/kubernetes/client-go/issues/193
- https://github.com/golang/go/issues/6213
- http://devel.io/2013/08/19/go-handling-arbitrary-json/

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/cpuguy83/go-md2man v1.0.10 // indirect
 	github.com/dchest/uniuri v0.0.0-20160212164326-8902c56451e9
 	github.com/directxman12/zapr v0.1.1
-	github.com/ghodss/yaml v1.0.0 // indirect
+	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/logr v0.1.0 // indirect
 	github.com/go-logr/zapr v0.1.0 // indirect
 	github.com/go-sql-driver/mysql v1.4.1 // indirect
@@ -57,12 +57,10 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 	k8s.io/api v0.0.0-20190222213804-5cb15d344471
-	k8s.io/apiextensions-apiserver v0.0.0-20190508191920-007dc40467c5 // indirect
 	k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
 	k8s.io/client-go v2.0.0-alpha.0.0.20190507014756-65905f29c17c+incompatible
 	k8s.io/klog v0.1.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20181024003938-96e8bb74ecdd // indirect
 	sigs.k8s.io/controller-runtime v0.1.10
-	sigs.k8s.io/testing_frameworks v0.1.1 // indirect
 	sigs.k8s.io/yaml v1.1.0 // indirect
 )
--- a/pkg/bosh/manifest/instance_group.go
+++ b/pkg/bosh/manifest/instance_group.go
@@ -3,6 +3,9 @@ package manifest
 import (
 	"fmt"
 
+	"github.com/ghodss/yaml"
+	corev1 "k8s.io/api/core/v1"
+
 	"code.cloudfoundry.org/cf-operator/pkg/kube/apis"
 )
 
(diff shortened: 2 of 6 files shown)


edwardstudy:
@viovanov I removed `wip` and it's ready for review. As I said, I changed `Manifest` struct tags to align kube resources struct for support affinity.

pkg/bosh/bpm/config.go
```diff
@@ -1,54 +1,54 @@
...
 
 import (
-	yaml "gopkg.in/yaml.v2"
+	"github.com/ghodss/yaml"
 )
 
 // Hooks from a BPM config
 type Hooks struct {
-	PreStart string `yaml:"pre_start,omitempty"`
+	PreStart string `json:"pre_start,omitempty"`
```
viovanov: This is a significant change. Assuming that `"github.com/ghodss/yaml"` needs json tags, you can still keep the yaml tags.

```
PreStart string `json:"pre_start,omitempty" yaml:"pre_start,omitempty"`
```

pkg/bosh/manifest/job.go
```diff
@@ -85,8 +88,58 @@ func (j *Job) sysDirs(name string) []string {
...
+	BOSHContainerization BOSHContainerization   `json:"bosh_containerization"  bson:"bosh_containerization"`
+	Properties           map[string]interface{} `json:"-" bson:",inline"`
+}
+
+// MarshalJSON This is a hack to support "inline" tag
+func (p *JobProperties) MarshalJSON() ([]byte, error) {
+	var i interface{}
+	b, err := bson.Marshal(p)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling 'JobProperties' into BSON: %v", err)
```
viovanov: This is not BSON - it's supposed to be JSON
What am I missing?

edwardstudy:
> This is a large change just for the ability to set affinity.
> I am particularly afraid of the `inline` "hack".
> 
> The Readiness/Liveness probes are marshaled/unmarshaled fine, can you please explain what problems you've found that caused you to make these changes?

Sure. 

1. First and foremost, `Affinity` struct has only `json` tags in upstream. So we could not decode `affinity` field in YAML. And I used `json` tags to decode directly. The second thing it that `inline` stuff you said. I agreed with you worry about the hack.

2. The second thing is what problme I saw. I followed Readiness/Liveness stuff to support `affinity` and [cases](https://github.com/cloudfoundry-incubator/cf-operator/pull/364/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension#diff-ff2b8e4c77044538210aebc4cc151552R534) failed because `YAML.Un marshal` could not work for following struct:
```
  env:
    bosh:
      agent:
        settings:
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                - matchExpressions:
                  - key: kubernetes.io/unit-test-az-name
                    operator: In
                    values:
                    - unit-test-az1
                    - unit-test-az2
```

And I noticed that `Readiness/Liveness` were not kube native struct: https://github.com/cloudfoundry-incubator/cf-operator/blob/e227241e67a1882857d6cefa23f73eeca45a8edb/testing/boshmanifest/manifests.go#L63-L72. The native kube `Readiness/Liveness` config should not have `handler`. So `affinity` struct maybe should do the same remakes that I don't want to do.

How's your opinion? we can talk in daily.

edwardstudy:
@viovanov Hi, I just created a branch to explain what happens when following `Readiness/Liveness` style

https://github.com/cloudfoundry-incubator/cf-operator/tree/ed/affinity-wip

Did not know why affinity is nil?

manno:
I talked to @viovanov  about that already and put my current idea in code: https://github.com/cloudfoundry-incubator/cf-operator/pull/378

@edwardstudy do you think something like that would help?
I would prefer a separate loader for the k8s native keys, so we don't have to change the BOSH structs.

I'm surprised we want to put native structs below `Env`, I thought that was limited to `BOSHContainerization`, but of course that's okay because `Env.bosh.agent.settings` is just "[Hash, optional]:" and we won't violate any schema.

Anyhow, I like the idea to have stronger separation between the 'BOSH input manifest', our additions and the output structs we use for k8s. I was a little confused that this is all in 'bosh/manifest' while working on the kube converter. But of course, we don't want to transform complicated structs just for the sake of it. And like Vlad mentioned, having Kube native structs is a much better fit for us and our users.

edwardstudy:
@manno Yes, It's more reasonable about this separation.I would like to enhance affinity part after we merge PR: https://github.com/cloudfoundry-incubator/cf-operator/pull/378 :)

edwardstudy:
@viovanov I totally changed this PR and made it clean.

manno:
Hey @edwardstudy I tried to use your branch with the split parser and it seems to work (well, if tests are green): https://github.com/cloudfoundry-incubator/cf-operator/pull/386

viovanov:
closed in favor of #386

PR #365, opened 2019-07-01
Modify debug logging for predicates and events

manno:
* make it less repetitive and more dense
* only log JSON to k8s event log
* only log text messages to debug log
* only pass one string
* more specialized methods

### DIFF
--- a/pkg/kube/util/ctxlog/events.go
+++ b/pkg/kube/util/ctxlog/events.go
@@ -13,9 +13,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-type event struct {
+type Event struct {
 	object runtime.Object
 	reason string
+}
+
+type PredicateEvent struct {
+	Event
+}
+
+type MappingEvent struct {
+	Event
 }
 
 const (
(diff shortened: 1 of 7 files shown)


pkg/kube/util/ctxlog/events.go
```diff
@@ -43,15 +47,49 @@ func WithEvent(object runtime.Object, reason string) EventLogger {
...
 
 	jsonData, _ := json.Marshal(objectInfo)
-	log.Debug(format, string(jsonData))
+	log.Debug(string(jsonData))
 
 	recorder := ExtractRecorder(ctx)
-	recorder.Event(ev.object, corev1.EventTypeNormal, ev.reason, fmt.Sprintf("%s %s", format, string(jsonData)))
+	var result map[string]string
+	json.Unmarshal([]byte(jsonData), &result)
+	recorder.Event(ev.object, corev1.EventTypeNormal, ev.reason, fmt.Sprintf("%s %s", msg, result["message"]))
```
qu1queee: @manno  from your comment in the PR
> I think the JSON should be in events, not in the debug log. Events only live for about an hour and are machine readable via the k8s AP

Why we are you not doing this now? I have the same opinion as you, keep the JSON in the events. Also, if the JSON will come here in events, the `msg` is not needed, either embedded inside the JSON, or remove it.

The rest of the code changes is nice, more compact, thanks!

PR #367, opened 2019-07-01
Change owner reference from job to extendedjob

rohitsakala:
[#166696566](https://www.pivotaltracker.com/n/projects/2192232/stories/166696566)


rohitsakala:
> @rohitsakala can you please explain what's different?
> I don't see an owner reference being set

My bad. I updated the commit and PR message. During our discussion regarding the fix for the bug [#166696566](https://www.pivotaltracker.com/n/projects/2192232/stories/166696566), you asked to put the owner reference of the versioned secret to the extended job and not a job. So, I have changed the owner reference for versioned secrets to the extended job which created it. Also, the bug persists if the owner reference is pointed to the job.

viovanov:
thanks @rohitsakala 
do the versioned secrets get properly deleted though?

rohitsakala:
> thanks @rohitsakala
> do the versioned secrets get properly deleted though?

Yes, with this PR the versioned secrets are deleted properly when I do `kubectl delete boshdeployment nats-deployment`. But when the owner reference was pointed to the job, the secrets were not getting deleted.

PR #373, opened 2019-07-02
[DO NOT MERGE] discussion notes

viovanov:


integration/deploy_test.go
```diff
@@ -291,6 +291,7 @@ var _ = Describe("Deploy", func() {
 					Expect(err).NotTo(HaveOccurred())
 
 					By("checking if the deployment was updated")
+					// TODO this does not ensure v2-0 is up or v2 is terminating while v3 is starting..
 					err = env.WaitForPod(env.Namespace, "test-nats-v2-1")
```
viovanov: It can't be terminating - we're looking for `running`, no?

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
```diff
@@ -113,10 +113,13 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
 
 	// Generate all the kube objects we need for the manifest
 	log.Debug(ctx, "Converting bosh manifest to kube objects")
+	// Use latest desired manifest version for dg/bpm jobs
+	// TODO start with invalid v0
 	version := r.resolver.LatestVersion(ctx, request.Namespace, manifest.Name)
```
viovanov: Why is this not ok? Because it starts with v1 even if it's not there?

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
```diff
@@ -130,6 +133,7 @@ func (r *ReconcileBOSHDeployment) Reconcile(request reconcile.Request) (reconcil
 	}
 
 	// Apply the "Data Gathering" ExtendedJob
+	// TODO `version` should be the version latest+1 created by `VariableInterpolationJob`, instead of hard coded
```
viovanov: why not `latest`, but `latest+1`?

pkg/kube/controllers/extendedjob/errand_controller.go
```diff
@@ -153,14 +153,16 @@ func AddErrand(ctx context.Context, config *config.Config, mgr manager.Manager)
 	}
 
 	// Watch secrets referenced by resource ExtendedJob
-	// trigger auto errand if UpdateOnConfigChange=true and config data changed
+	// trigger auto errand if UpdateOnConfigChange=true and config data changed.
+	// This includes changes to any version of a referenced 'versioned secret'.
```
viovanov: They should be immutable..
If possible, we should ignore "changes" to any secret that is versioned.
We should only care about creation.

pkg/kube/controllers/extendedjob/errand_reconciler.go
```diff
@@ -117,6 +117,8 @@ func (r *ErrandReconciler) createJob(ctx context.Context, eJob ejv1.ExtendedJob)
 	}
 	template.Labels[ejv1.LabelEJobName] = eJob.Name
 
+	// TODO modify secret references to use latest version of 'versioned secret' for the job
+	// TODO don't create a job if any versioned secret still has v0
```
viovanov: +1 for using latest
What does delaying job creation get us?
Won't we create for v1 when it pops up anyway?

PR #375, opened 2019-07-03
Make all test use OPERATOR_TEST_STORAGE_CLASS env variable

rohitsakala:
[#166924183](https://www.pivotaltracker.com/n/projects/2192232/stories/166924183)


docs/examples/extended-statefulset/exstatefulset_pvcs.yaml
```diff
@@ -30,4 +30,4 @@ spec:
           accessModes: [ "ReadWriteOnce" ]
           resources:
             requests:
-              storage: 1Gi              
+              storage: 1Gi
```
edwardstudy: newline

e2e/kube/examples_test.go
```diff
@@ -159,9 +189,17 @@ var _ = Describe("Examples", func() {
 			It("bosh-deployment with a persistent disk example must work", func() {
 				yamlFilePath := examplesDir + "bosh-deployment/boshdeployment-with-persistent-disk.yaml"
 
+				By("Creating a secret for implicit variable")
```
edwardstudy: variable?

integration/environment/machine.go
```diff
@@ -212,7 +212,7 @@ func (m *Machine) PVCsDeleted(namespace string) (bool, error) {
 
 // WaitForStatefulSet blocks until all statefulset pods are running. It fails after the timeout.
 func (m *Machine) WaitForStatefulSet(namespace string, labels string) error {
-	return wait.PollImmediate(m.pollInterval, m.pollTimeout, func() (bool, error) {
+	return wait.PollImmediate(m.pollInterval, 500*time.Second, func() (bool, error) {
```
edwardstudy: Why not increase `m.pollTimeout`? Is this timeout special?

viovanov: Why wait 8 minutes? This is a bad idea.
We should redesign our tests if something like this is required.

rohitsakala: I was just testing this flaky test. I added a wip tag too.

docs/examples/extended-statefulset/exstatefulset_pvcs.yaml
rohitsakala: Done.

e2e/kube/examples_test.go
rohitsakala: The `bosh-deployment-persistent-disk.yaml` example uses an implicit variable to specify the storage class for the persistent disk. For the implicit variable, I am creating a secret to specify the value of the storage class. (https://github.com/cloudfoundry-incubator/cf-operator/blob/rohit/storageclass/docs/examples/bosh-deployment/boshdeployment-with-persistent-disk.yaml#L21)

integration/environment/machine.go
viovanov: Ah, I got scared :P

PR #378, opened 2019-07-04
Extract BOSH containerization YAML loader 

manno:
Our deployment manifest contains structures from both BOSH and Kubernetes.

* We want it to be a valid BOSH manifest
* We want to re-use structures, which we add for to the BOSH manifest for containerization, without transforming them to k8s native structs.

This introduces a separate YAML loader for `BOSHContainerization` and `env.bosh.agent.settings.affinity` keys.

### DIFF
--- a/pkg/bosh/manifest/containerization/containerization.go
+++ b/pkg/bosh/manifest/containerization/containerization.go
@@ -14,15 +14,38 @@ import (
 type Manifest struct {
 	InstanceGroups []*InstanceGroup `yaml:"instance_groups,omitempty"`
 }
+
 type InstanceGroup struct {
 	Jobs []Job `yaml:"jobs"`
+	// for env.bosh.agent.settings.affinity
+	Env Env `yaml:"env,omitempty"`
 }
+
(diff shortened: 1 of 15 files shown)


pkg/bosh/manifest/containerization/containerization.go
```diff
@@ -1,13 +1,55 @@
...
 import (
+	yaml "gopkg.in/yaml.v2"
+
 	corev1 "k8s.io/api/core/v1"
 
 	"code.cloudfoundry.org/cf-operator/pkg/bosh/bpm"
 )
 
+// Manifest is a BOSH deployment manifest
+type Manifest struct {
```
edwardstudy: Why these structs like Manifest, InstanceGroup should be containerization-specified? I think it should be bosh part.

manno: I thought I need those structs so we can get to the affinity and bosh_containerization keys in the tree.

viovanov:
closed in favor of #386

PR #380, opened 2019-07-04
Improved image build

f0rmiga:
It builds the image on the host, then copies the final binary to the
image. It is useful while leveraging the host cache during the build
process.


f0rmiga:
Closing this as we decided against it right now due to the need of extra dependencies in the host.

PR #381, opened 2019-07-04
Enabled separate namespace for deployment and operation

f0rmiga:
The namespace that cf-operator is deployed can be separate from the one that it operates on. This commit enables it if `env.CF_OPERATOR_NAMESPACE` is specified in the Helm installation.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
 export OPERATOR_DOCKER_ORGANIZATION=${OPERATOR_DOCKER_ORGANIZATION:-'cfcontainerization'}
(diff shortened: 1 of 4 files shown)


qu1queee:
Have the cf-operator enough permissions to read resources in other namespaces?

f0rmiga:
@qu1queee I was able to. And RBAC is enabled:
```
$ kubectl api-versions | grep rbac
rbac.authorization.k8s.io/v1
rbac.authorization.k8s.io/v1beta1
```

f0rmiga:
Closing now as it has not been discussed so far. Going to discuss in the future.

PR #382, opened 2019-07-04
Added bin/helm-template

f0rmiga:
It runs helm-template and then pipes it to kubectl apply or delete.


f0rmiga:
Closing this as it is not relevant for now.

PR #383, opened 2019-07-05
Add a sidecar that tails all logs in /var/vcap/sys/log

viovanov:


f0rmiga:
@viovanov LGTM. Just need to update the unit tests.

PR #390, opened 2019-07-07
Make all test use OPERATOR_TEST_STORAGE_CLASS env variable

rohitsakala:
[#166924183](https://www.pivotaltracker.com/n/projects/2192232/stories/166924183)


testing/kubectl_helper.go
```diff
@@ -292,3 +307,28 @@ func (k *Kubectl) SecretCheckData(namespace string, secretName string, fieldPath
...
+	extendedStatefulSetBytes, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return err
+	}
+	err = yaml.Unmarshal(extendedStatefulSetBytes, &extendedStatefulSet)
+	if err != nil {
+		return err
+	}
+
+	extendedStatefulSet.Spec.Template.Spec.VolumeClaimTemplates[0].Spec.StorageClassName = util.String(class)
```
edwardstudy: Could we check the existence of properties firstly to avoid NPE?

e2e/kube/storage/examples_storage_test.go
```diff
@@ -34,6 +47,36 @@ var _ = Describe("Examples", func() {
...
 			})
+
+			It("extended-statefulset pvc example must work", func() {
+				yamlFilePath := examplesDir + "extended-statefulset/exstatefulset_pvcs.yaml"
+
+				// Adding storageclass to volumeclaimtemplates
+				class, ok := os.LookupEnv("OPERATOR_TEST_STORAGE_CLASS")
+				Expect(ok).To(Equal(true))
+
+				err := kubectlHelper.AddTestStorageClassToVolumeClaimTemplates(yamlFilePath, class)
```
aduffeck: If I read the code correctly this actually alters the file in the `exampes` directory, depending on the environment variable. I think this file shouldn't be changed.

e2e/kube/storage/examples_storage_test.go
```diff
@@ -9,18 +11,29 @@ import (
 
 var _ = Describe("Examples", func() {
 
-	Describe("when examples are specified in the docs", func() {
+	Describe("when storage related examples are specified in the docs", func() {
 
+		var (
+			kubectlHelper *testing.Kubectl
+		)
+		kubectlHelper = testing.NewKubectl()
```
aduffeck: This should go into a `BeforeEach` block.

e2e/kube/storage/examples_storage_test.go
```diff
@@ -9,18 +11,29 @@ import (
...
 
+				By("Creating a secret for implicit variable")
+				class, ok := os.LookupEnv("OPERATOR_TEST_STORAGE_CLASS")
+				Expect(ok).To(Equal(true))
+				literalValues := "--from-literal=value=" + class
+				err := kubectlHelper.CreateSecretFromLiteral(namespace, "nats-deployment.var-implicit-operator-test-storage-class", "generic", literalValues)
+				Expect(err).ToNot(HaveOccurred())
+
 				By("Creating bosh deployment")
 				kubectlHelper := testing.NewKubectl()
```
aduffeck: I think you meant to extract this into a context variable in https://github.com/cloudfoundry-incubator/cf-operator/pull/390/commits/fcb4e0dc002dcec551f94a2fc157c408095aff56#diff-fdb6df78666e4d68c918c8133e965eb7R17, so this line can go away.

e2e/kube/storage/examples_storage_test.go
```diff
@@ -34,6 +47,36 @@ var _ = Describe("Examples", func() {
...
+
+				// Adding storageclass to volumeclaimtemplates
+				class, ok := os.LookupEnv("OPERATOR_TEST_STORAGE_CLASS")
+				Expect(ok).To(Equal(true))
+
+				err := kubectlHelper.AddTestStorageClassToVolumeClaimTemplates(yamlFilePath, class)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("Creating exstatefulset pvcs")
+				kubectlHelper := testing.NewKubectl()
```
aduffeck: Again, this line is not needed anymore.

integration/storage/extended_statefulset_with_storage_test.go
```diff
@@ -35,7 +36,9 @@ var _ = Describe("ExtendedStatefulSet", func() {
 		BeforeEach(func() {
 
 			essName := fmt.Sprintf("testess-%s", helper.RandString(5))
-			extendedStatefulSet = env.ExtendedStatefulSetWithPVC(essName, "pvc")
+			storageClass, ok := os.LookupEnv("OPERATOR_TEST_STORAGE_CLASS")
+			Expect(ok).To(Equal(true))
```
aduffeck: It's just syntactic sugar, but there's also `BeTrue()`.

testing/kubectl_helper.go
```diff
@@ -59,6 +63,16 @@ func (k *Kubectl) Create(namespace string, yamlFilePath string) error {
 	return nil
 }
 
+// CreateSecretFromLiteral creates a generic type secret using kubectl command
+func (k *Kubectl) CreateSecretFromLiteral(namespace string, secretName string, secretType string, literalValues string) error {
```
aduffeck: I think this only works for `generic` secrets anyway, no? So this parameter could be omitted.

`literalValues` implies that it's possible to set multiple values while it's only possible to set one key-value pair afaik. I'd change it to singular.

I think it's also better to just take the payload as an parameter here and not expect the caller to add the `--from-literal=` part.

e2e/kube/storage/examples_storage_test.go
rohitsakala: Done.

e2e/kube/storage/examples_storage_test.go
rohitsakala: Done.

e2e/kube/storage/examples_storage_test.go
rohitsakala: Done.

e2e/kube/storage/examples_storage_test.go
rohitsakala: Done.

integration/storage/extended_statefulset_with_storage_test.go
rohitsakala: Done.

testing/kubectl_helper.go
rohitsakala: Yup. Gotcha. Done

PR #394, opened 2019-07-08
Update development havener file

qu1queee:
Add support for ENV variables
Remove common prefix `chart` from releases list. While this format have been deprecated in the latest release.

This is a no-op, please merge.

### DIFF
--- a/dev-env-havener.yaml
+++ b/dev-env-havener.yaml
@@ -29,16 +29,22 @@
 
 name: cf-operator-quarks
 releases:
-- chart_name: cf-operator
-  chart_namespace: quarks
-  chart_location: deploy/helm/cf-operator/
+- name: cf-operator
+  namespace: quarks
+  location: deploy/helm/cf-operator/
   overrides:
     image:
-      tag: (( shell . bin/include/versioning && echo "${VERSION_TAG}" ))
-      repository: cf-operator
-      org: cfcontainerization
+      tag: (( env DOCKER_IMAGE_TAG ))
+      repository: (( env DOCKER_REPOSITORY ))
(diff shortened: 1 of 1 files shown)


qu1queee:
@viovanov can we merge?

dev-env-havener.yaml
```diff
@@ -29,16 +29,22 @@
...
-  chart_location: deploy/helm/cf-operator/
+- name: cf-operator
+  namespace: quarks
+  location: deploy/helm/cf-operator/
   overrides:
     image:
-      tag: (( shell . bin/include/versioning && echo "${VERSION_TAG}" )) 
-      repository: cf-operator
-      org: cfcontainerization
+      tag: (( env IMAGE_TAG ))
```
manno: we already have `DOCKER_IMAGE_TAG`, can we use that here too?

dev-env-havener.yaml
```diff
@@ -29,16 +29,22 @@
...
+      repository: (( env DOCKER_REPOSITORY ))
+      org: (( env DOCKER_ORGANIZATION ))
     customResources:
-      enableInstallation: false
+      enableInstallation: (( env CRDS_ENABLED ))
+
+env:
+  IMAGE_TAG: (( shell . bin/include/versioning && echo "${VERSION_TAG}" ))
```
manno:
```suggestion
  DOCKER_IMAGE_TAG: (( shell . bin/include/versioning && echo "${VERSION_TAG}" ))
```

dev-env-havener.yaml
qu1queee: yes

PR #396, opened 2019-07-08
Separate storage tests for e2e and integration

rohitsakala:
[#167098433](https://www.pivotaltracker.com/n/projects/2192232/stories/167098433)

CI related PR - https://github.com/cloudfoundry-incubator/cf-operator-ci/pull/66


bin/test-helm-e2e
```diff
@@ -9,19 +9,17 @@ if [ -z ${DOCKER_IMAGE_TAG+x} ]; then
   export DOCKER_IMAGE_TAG
 fi
 
-if [ -z ${TEST_NAMESPACE+x} ]; then
```
manno: Is that necessary? I like to export a fixed namespace when testing, so I don't have to change my `kubectl` statements every time...

rohitsakala: Yup. Reverted the change.

PR #400, opened 2019-07-08
Add the ability to freeze failing prestart hook init containers

viovanov:


pkg/bosh/manifest/container_factory.go
```diff
@@ -342,8 +344,14 @@ func boshPreStartInitContainer(
 	jobName string,
 	jobImage string,
 	volumeMounts []corev1.VolumeMount,
+	debug bool,
 ) corev1.Container {
 	boshPreStart := filepath.Join(VolumeJobsDirMountPath, jobName, "bin", "pre-start")
+	debugScript := ""
```
f0rmiga: For clarity, I'd suggest:

```
var script string
if debug {
    script = fmt.Sprintf(`if [ -x "%[1]s" ]; then "%[1]s"; fi`, boshPreStart)
} else {
    script = fmt.Sprintf(`if [ -x "%[1]s" ]; then "%[1]s" || ( echo "Debug window 1hr" ; sleep 3600 ); fi`, boshPreStart)
}
```

pkg/bosh/manifest/container_factory.go
```diff
@@ -362,7 +370,13 @@ func bpmPreStartInitContainer(
 	process bpm.Process,
 	jobImage string,
 	volumeMounts []corev1.VolumeMount,
+	debug bool,
 ) corev1.Container {
+	debugScript := ""
+	if debug {
+		debugScript = ` || ( echo "Debug window 1hr" ; sleep 3600 ) `
```
f0rmiga: Same here.

pkg/bosh/manifest/container_factory.go
```diff
@@ -362,7 +373,16 @@ func bpmPreStartInitContainer(
 	process bpm.Process,
 	jobImage string,
 	volumeMounts []corev1.VolumeMount,
+	debug bool,
 ) corev1.Container {
+
+	var script string
+	if debug {
+		script = fmt.Sprintf(`%s || ( echo "Debug window 1hr" ; sleep 3600`, process.Hooks.PreStart)
```
f0rmiga: This one is missing closing the subshell.

PR #402, opened 2019-07-09
Create or Update Twice

manno:
Deployment reconciler doesn't reset errand state

Errands will get triggers from changes anyhow.
Errand no longer executes twice when updating the deployment.
Deployment no longer jumps from v2 to v4, but to v3 as expected.

Also:
* wrapping the reconcile error before returning it
* not requeuing failed updates to state
* clean up to manifest resolver
* add job output into debug log

[#167037095](https://www.pivotaltracker.com/story/show/167037095)


edwardstudy:
I love this PR about the desired manifest. So I merged it for [#167098712](https://www.pivotaltracker.com/story/show/167098712) and changed other `controllerutil.CreateOrUpdate` funcs.

PR #404, opened 2019-07-10
Handle issues when operator restarts

edwardstudy:
[#167098712](https://www.pivotaltracker.com/story/show/167098712)

- desired manifest would not be updated if not changes
- only process create events of ESTS which have no children resources
- only process create events of ExtendedSecret which have no generated secret

### DIFF
--- a/pkg/kube/controllers/extendedstatefulset/util.go
+++ b/pkg/kube/controllers/extendedstatefulset/util.go
@@ -8,7 +8,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	crc "sigs.k8s.io/controller-runtime/pkg/client"
 
-	bdm "code.cloudfoundry.org/cf-operator/pkg/bosh/manifest"
 	estsv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/extendedstatefulset/v1alpha1"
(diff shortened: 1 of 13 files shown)


edwardstudy:
Manual verified. And will add e2e cases for next story.

pkg/kube/controllers/extendedstatefulset/util.go
```diff
@@ -0,0 +1,44 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	crc "sigs.k8s.io/controller-runtime/pkg/client"
+
+	estsv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/extendedstatefulset/v1alpha1"
+	"code.cloudfoundry.org/cf-operator/pkg/kube/util/ctxlog"
+)
+
+// listStatefulSets gets all StatefulSets cross version owned by the ExtendedStatefulSet
+func listStatefulSets(ctx context.Context, client crc.Client, exStatefulSet *estsv1.ExtendedStatefulSet) ([]v1beta2.StatefulSet, error) {
```
manno: @edwardstudy  I'm wondering if we can put this one and `listSecrets` into the `pkg/kube/apis/extended*/*/types.go` files, or if that would create an import cycle.

PR #406, opened 2019-07-10
Modify the webhook validator for BOSHDeployment type

qu1queee:
[#167145276](https://www.pivotaltracker.com/n/projects/2192232/stories/167145276)

- Do not allow admissions if the referenced spec.ops resources are not in place.

- For the spec.ops resources availability
  - Allow to check during 5 secs, if the resource is there(timeout after)
  - Only supported for configmaps and secrets

- Added integration tests that cover
  - not existing spec.ops resource referenced in the BOSHDeployment
  - use case when the spec.ops resource takes time to come up, but still under
    the timeout limits
  - use case when the spec.ops resource takes time to come up, but hits the
    timeout

### DIFF
--- a/pkg/kube/controllers/boshdeployment/validating_webhook.go
+++ b/pkg/kube/controllers/boshdeployment/validating_webhook.go
@@ -59,10 +59,12 @@ func AddBOSHDeploymentValidator(log *zap.SugaredLogger, config *config.Config, m
 
 // Validator represents a validator for BOSHDeployments
 type Validator struct {
-	log     *zap.SugaredLogger
-	config  *config.Config
-	client  client.Client
-	decoder types.Decoder
+	log          *zap.SugaredLogger
+	config       *config.Config
+	client       client.Client
+	decoder      types.Decoder
+	pollTimeout  time.Duration
+	pollInterval time.Duration
 }
 
 // NewValidator returns a new BOSHDeploymentValidator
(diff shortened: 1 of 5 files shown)


testing/catalog.go
```diff
@@ -815,6 +815,19 @@ func (c *Catalog) DefaultBOSHDeploymentWithOps(name, manifestRef string, opsRef
 	}
 }
 
+// DefaultBOSHDeploymentWithUnsupportedOps fissile deployment CR with ops
```
edwardstudy: I don't know why there is fissile deployment? do we have any history about it?

pkg/kube/controllers/boshdeployment/validating_webhook.go
```diff
@@ -69,6 +76,34 @@ func NewValidator(log *zap.SugaredLogger, config *config.Config) admission.Handl
 	}
 }
 
+// OpsResourceExist verify if a resource exist in the namespace,
+// it will check it´s existance during 5 seconds,
+// otherwise it will timeout.
+func (v *Validator) OpsResourceExist(ctx context.Context, specOpsResource bdv1.Ops, ns string) (bool, string) {
+	timeOut := time.After(5 * time.Second)
```
edwardstudy: Could we reuse these values or parameterize them:

https://github.com/cloudfoundry-incubator/cf-operator/blob/38b100de0943e09ce2b82ce611055dbdd095f120/integration/environment/machine.go#L31-L32

testing/catalog.go
qu1queee: I think it was a plain copy paste of a existing funcs, will take a look

pkg/kube/controllers/boshdeployment/validating_webhook.go
qu1queee: yeah, thanks!

qu1queee:
This have a bug, is a ginkgo issue, while I run some specs in goroutines. Trying to find out the root cause.

PR #412, opened 2019-07-11
Enable remove_namespaces in CI

rohitsakala:


manno:
Ok we have the re-ordering in https://github.com/cloudfoundry-incubator/cf-operator/pull/413

I suggested removing the namespace deletion from the scripts, as it is done by the suite and could be added to CI: https://github.com/cloudfoundry-incubator/cf-operator/pull/414

rohitsakala:
> Ok we have the re-ordering in #413
> 
> I suggested removing the namespace deletion from the scripts, as it is done by the suite and could be added to CI: #414

Ok. thanks.

PR #415, opened 2019-07-11
Fix go lint warnings

manno:


pkg/kube/controllers/boshdeployment/bpm_reconciler.go
```diff
@@ -6,7 +6,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
```
edwardstudy: It confuses me for a long time. My editor says that redundant alias about this import.

manno: My editor forces that every time... 
I think it's goimports: https://github.com/golang/go/issues/26882#issuecomment-506549469
and https://github.com/golang/go/issues/29556

manno: I removed goimports from my setup.

PR #420, opened 2019-07-15
Add a restart case for e2e test

edwardstudy:
[#167102529](https://www.pivotaltracker.com/story/show/167102529)

- Determined if EJob needs to updated for deployment to fix duplicated EJob triggers
- Determined if ESTS needs to updated for deployment
- Added `generated` status to control secret generations
- Listed StatefulSets from API client directly for ESTS Create event process


pkg/kube/controllers/boshdeployment/deployment_reconciler.go
```diff
@@ -209,9 +209,11 @@ func (r *ReconcileBOSHDeployment) createEJob(ctx context.Context, instance *bdv1
 
 	op, err := controllerutil.CreateOrUpdate(ctx, r.client, eJob.DeepCopy(), func(obj runtime.Object) error {
 		if existingEJob, ok := obj.(*ejv1.ExtendedJob); ok {
-			eJob.ObjectMeta.ResourceVersion = existingEJob.ObjectMeta.ResourceVersion
-			eJob.Spec.Trigger.Strategy = existingEJob.Spec.Trigger.Strategy
-			eJob.DeepCopyInto(existingEJob)
+			if shouldEJobUpdate(existingEJob, eJob) {
```
manno: `CreateOrUpdate` should be a no-op if the ejob exists and hasn't changed. We copy strategy and resource version to avoid unnecessary updates. So I guess there was a case where we did update the ejobs, even though we didn't have to?

edwardstudy: @manno Yes. After we copy strategy and resource version, CreateOrUpdate still update EJob.

edwardstudy:
Still have randomly update ESTS /sad

edwardstudy:
I can get empty list after restarting operator

https://github.com/cloudfoundry-incubator/cf-operator/blob/4b64fe2fc87c59565b25f58843ca49c6bb9430f1/pkg/kube/controllers/extendedstatefulset/util.go#L21-L32

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
manno: Seems like we have two strategies to avoid the unwanted update now. `shouldEJobUpdate` and the copying of the previous attributes.

Also I wasn't able to see CreateOrUpdate misbehave before in https://www.pivotaltracker.com/story/show/167037095

edwardstudy: We can see this action when we restart operator which mean that CreateOrUpdate did update for EJob resource:
```
2019-07-19T12:00:13.502+0800	DEBUG	boshdeployment-reconciler	boshdeployment/deployment_reconciler.go:212  	ExtendedJob 'bpm-nats-deployment' has been updated
```
IIRC, we didn't get operation results 10 days ago.

PR #422, opened 2019-07-16
Fix implicit variable processing

viovanov:


pkg/bosh/manifest/resolver.go
```diff
@@ -81,17 +79,6 @@ func (r *Resolver) WithOpsManifest(instance *bdc.BOSHDeployment, namespace strin
 		return manifest, errors.Wrapf(err, "failed to unmarshal manifest")
 	}
 
-	// Interpolate implicit variables
-	vars := r.implicitVariables(manifest, m)
-	for _, v := range vars {
-		varData, err := r.resourceData(namespace, bdc.SecretType, names.CalculateSecretName(names.DeploymentSecretTypeImplicitVariable, instance.GetName(), v), bdc.ImplicitVariableKeyName)
```
manno: It should be safe to remove the bdc.SecretType case from the `func resourceData` now.

PR #423, opened 2019-07-16
Add log-level command line switch

manno:
* `--log-level=` controls minimal log level
* prefix controller-runtime logs with 'cr',
  that seems to result in 'cr.kubebuilder' prefixed log entries.
* configure caller skip for predicate related logging, so we see
  the actual caller location in the logs

[#167086778](https://www.pivotaltracker.com/story/show/167086778)


pkg/kube/util/ctxlog/context.go
```diff
@@ -34,6 +34,11 @@ func NewContextWithRecorder(ctx context.Context, name string, recorder record.Ev
 	return context.WithValue(ctx, ctxLoggerKey, log)
 }
 
+func ExtractLoggerWithSkip(ctx context.Context, skip int) *zap.SugaredLogger {
```
edwardstudy: We should give a comment if we follow go lint.

cmd/internal/root.go
```diff
@@ -132,14 +137,20 @@ func init() {
 
 // newLogger returns a new zap logger
 func newLogger(options ...zap.Option) *zap.SugaredLogger {
-
-	logger, err := zap.NewDevelopment(options...)
+	level := viper.GetString("log-level")
+	l := zap.DebugLevel
+	l.Set(level)
+
+	cfg := zap.NewDevelopmentConfig()
```
edwardstudy: Why we don't use `NewProductionConfig` directly if we don't want to keep development env? Do we have customized something?

manno: It's very different from the log format we have now: https://github.com/uber-go/zap/blob/master/config.go#L114-L127
Judging from the code it outputs everything as JSON.
I tried it a few months back and if I remember correctly, it doesn't contain code locations.

Setting `Development: false` just makes the stack traces more dense, according to documentation.

pkg/kube/util/ctxlog/context.go
manno: Yes, even worse, the function was ignoring the skip parameter....

PR #430, opened 2019-07-17
Persisting output changes for dg and bpm

qu1queee:
[#167113640](https://www.pivotaltracker.com/story/show/167113640)

For the job_reconciler

- Do not print container logs if the cmd call fails
- Generate ExtendedJob event, when the JSON unmarshalling fails, this helps
  to understand, that a secret will be not be created, and why the pod did not came up.

For the bpm-configs and data-gather commands

- Ensure that when writing to stdOut, only the JSON bytes are there. To
  approach this issue, save the original stdOut, and write the cmd main
  logic into a temporal stdOut. Prior to dumping the JSON bytes into
  stdOut, we can reset to the original stdOut.
  When changing the stdOut to a temporal one, we do save the temporal state in a
  buffer, and later check if there is something there. This helps to ensure, that in case 
  of errors, we do not lose the data.
- Because when we dump the JSON into a buffer, the `bufio.NewWriter` allows us to decide
  when to dump that data into disk, this via the `Flush()`. Because we `defer` the flush, we can   ensure that always our JSON  is at the very end of the stdOut. Therefore, when reading logs from the pods that generate the persistenOutput, we can read them using `tail`, to ensure that in case of errors, we just process the JSON piece.


cmd/internal/bpm_configs.go
```diff
@@ -25,71 +27,103 @@ This command calculates and prints the BPM configurations for all all BOSH jobs
...
-
-		f := bufio.NewWriter(os.Stdout)
-		defer f.Flush()
-		_, err = f.Write(jsonBytes)
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return gatherInstanceGroupBPMConfigs()
```
manno: I'm not a huge fan of this private function. We're not going to call it from somewhere else, it doesn't help documenting the code. Also it seems to break the 'cobra' style, which we're using in all the other commands.

We're already passing a function into the struct: `RunE: func(cmd *cobra.Command, args []string) error {`

cmd/internal/bpm_configs.go
```diff
@@ -25,71 +27,103 @@ This command calculates and prints the BPM configurations for all all BOSH jobs
...
+	os.Stdout = origStdOut
+
+	var buf bytes.Buffer
+	io.Copy(&buf, r)
+
+	if buf.Len() > 0 {
+		log.Debugf("Unexpected data on the data gathering, dumping from buffer: %s", buf.String())
+	}
+
+	// Write to an original stdOut
```
manno: These comments have line breaks very early?

cmd/internal/bpm_configs.go
qu1queee: I will revert.

cmd/internal/bpm_configs.go
```diff
@@ -79,7 +90,25 @@ instance group.
...
 
+		// Close w, and restore the original stdOut
+		w.Close()
+		os.Stdout = origStdOut
+
+		var buf bytes.Buffer
+		io.Copy(&buf, r)
+
+		if buf.Len() > 0 {
+			return errors.Errorf("unexpected data sent to std=ut, during the data bpm-configs cmd: %s", buf.String())
```
manno:
```suggestion
			return errors.Errorf("unexpected data sent to stdout, during the data bpm-configs cmd: %s", buf.String())
```

cmd/internal/dgather.go
```diff
@@ -2,17 +2,18 @@ package cmd
...
 	"io/ioutil"
 	"os"
 
+	"code.cloudfoundry.org/cf-operator/pkg/bosh/manifest"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	yaml "gopkg.in/yaml.v2"
-
-	"code.cloudfoundry.org/cf-operator/pkg/bosh/manifest"
```
manno: While you're at it, can you move that back to line 15? 
It's only style, but we group import statements like go, github, k8s, our code in a lot of files.

integration/environment/machine.go
```diff
@@ -5,25 +5,23 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
-
-	bdv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/boshdeployment/v1alpha1"
-	ejv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/extendedjob/v1alpha1"
-	esv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/extendedsecret/v1alpha1"
-	essv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/extendedstatefulset/v1alpha1"
-	"code.cloudfoundry.org/cf-operator/pkg/kube/client/clientset/versioned"
-	"code.cloudfoundry.org/cf-operator/pkg/kube/util"
-	podutil "code.cloudfoundry.org/cf-operator/pkg/kube/util/pod"
```
manno: again import statement grouping

pkg/kube/controllers/extendedjob/job_reconciler.go
```diff
@@ -178,24 +180,28 @@ func (r *ReconcileJob) persistOutput(ctx context.Context, instance *batchv1.Job,
...
-		ctxlog.Debug(ctx, string(result))
+
+		// Create secret
+		secretName := ejob.Spec.Output.NamePrefix + c.Name
 
 		var data map[string]string
 		err = json.Unmarshal(result, &data)
 		if err != nil {
+			ctxlog.WithEvent(&ejob, "ExtendedJob").Errorf(ctx, "non-JSON output was emitted for container '%s', secret '%s' cannot be created", instance.GetName(), secretName)
 			return errors.Wrap(err, "invalid output format")
```
manno: I think `Errorf` returns an err, I would just use that one here, so we have the same message.
Also the line above, it could be invalid JSON, not just non-JSON ;)

Hm, when `persistOutput` fails we create another error.

pkg/kube/controllers/extendedjob/job_reconciler.go
```diff
@@ -178,24 +180,28 @@ func (r *ReconcileJob) persistOutput(ctx context.Context, instance *batchv1.Job,
...
 		}
 
-		// Create secret and persist the output
-		secretName := ejob.Spec.Output.NamePrefix + c.Name
+		// Put full job output in debug log
+		ctxlog.Debugf(ctx, "Printing logs from container %s: %s", c.Name, string(result))
```
manno: We talked about that in the daily, can you remove that, while you're at it? Sorry :)

integration/extended_job_test.go
```diff
@@ -4,14 +4,12 @@ import (
...
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-
 	"code.cloudfoundry.org/cf-operator/integration/environment"
 	ejv1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/extendedjob/v1alpha1"
 	"code.cloudfoundry.org/cf-operator/pkg/kube/util"
 	"code.cloudfoundry.org/cf-operator/pkg/kube/util/versionedsecretstore"
-
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
manno: import style

cmd/internal/bpm_configs.go
qu1queee: fixed

cmd/internal/dgather.go
qu1queee: done

pkg/kube/controllers/extendedjob/job_reconciler.go
qu1queee: thanks

PR #432, opened 2019-07-17
Add addon support

viovanov:


pkg/bosh/manifest/addon_placement.go
```diff
@@ -0,0 +1,95 @@
...
+	return false, nil
+}
+
+// jobMatch matches job rules for addon placement
+func (m *Manifest) jobMatch(instanceGroup *InstanceGroup, rules *AddOnPlacementRules) (bool, error) {
+	if instanceGroup == nil || rules == nil {
+		return false, nil
+	}
+
+	jobList := map[string]bool{}
```
f0rmiga: [nit] It's better to use `map[string]struct{}`.

pkg/bosh/manifest/resolver.go
```diff
@@ -86,47 +86,53 @@ func (r *Resolver) WithOpsManifest(instance *bdc.BOSHDeployment, namespace strin
...
-			return manifest, errors.Wrapf(err, "failed to interpolate %#v", m)
+			return nil, errors.Wrapf(err, "failed to interpolate %#v", m)
 		}
 	}
 
 	// Reload the manifest after interpolation, and apply implicit variables
 	manifest, err = LoadYAML(bytes)
 	if err != nil {
-		return manifest, errors.Wrapf(err, "failed to load yaml after applying ops %#v", m)
+		return nil, errors.Wrapf(err, "failed to load yaml after applying ops %#v", m)
```
f0rmiga: Thanks for returning nil. :)

PR #434, opened 2019-07-18
Fix addon links and improve debuggability

viovanov:


pkg/bosh/manifest/data_gatherer.go
```diff
@@ -69,6 +69,13 @@ func (jpl JobProviderLinks) Add(job Job, spec JobSpec, jobsInstances []bc.JobIns
 
 		if providers, ok := jpl[linkType]; ok {
 			if _, ok := providers[linkName]; ok {
+
```
edwardstudy: better remove this empty lin

cmd/internal/dgather.go
```diff
@@ -25,7 +26,12 @@ This will retrieve information of an instance-group
 inside a bosh manifest file.
 
 `,
-	RunE: func(cmd *cobra.Command, args []string) error {
+	RunE: func(cmd *cobra.Command, args []string) (err error) {
+		defer func() {
+			if err != nil {
+				time.Sleep(time.Second * 5)
```
manno: Maybe put that in a constant and explain in a comment why the delay is there.

PR #437, opened 2019-07-18
Changed the flow image to a newly layouted one.

andreas-kupries:
I found the existing diagram of the rendering flow difficult to follow.
I made a copy of the original and then re-layouted it to hopefully be
easier to follow. The new image shows the flow from top to bottom, in
layers, with each layer only depending on the previous layers.


manno:
I created https://www.draw.io/#G13op-0R9h3ZLmjFWDwC4DTXPwn7jjEG1- a while ago, which focuses more on updates. Should we also add that?

PR #441, opened 2019-07-19
Use DOCKER_IMAGE_ORG instead of OPERATOR_DOCKER_ORGANIZATION

manno:
Also remove useless curly braces and quote the version var...

[#167388804](https://www.pivotaltracker.com/story/show/167388804)


bin/build-image
```diff
@@ -3,25 +3,25 @@
 set -o errexit
 
 GIT_ROOT="${GIT_ROOT:-$(git rev-parse --show-toplevel)}"
-. "${GIT_ROOT}/bin/include/versioning"
```
edwardstudy: The reason for changes is by some script check tool?

manno: No, just removing useless expansion brackets. They're not recommended by shellcheck and I think they give a false sense of security and add line noise. The `${GIT_ROOT}` is not using variable expansion features like in the line above.

This line from the PR for example:
`DOCKER_IMAGE_TAG=${VERSION_TAG}` is identical to `DOCKER_IMAGE_TAG=$VERSION_TAG` and both are actually protected from pathname expansion and word splitting. However I still would add double quotes to be consistent: `DOCKER_IMAGE_TAG="$VERSION_TAG"`.

PR #454, opened 2019-07-24
cf-operator util tail-logs cmd

qu1queee:
Add tail-logs subcommand

This `util` cmd is intended to run as a container command, for tailing
logs of a parallel container at an specific directory.

This subcommand will do the following:
- tail for files under an specified dir
- tail only files of the type "*.log"
- stream files content to stdout of the container where it runs
- add colors per file, when running `kubectl logs` , each different file
  will be displayed with an specific color

This pr also includes:
- new integration suite to test subcommands, this should help for future
  subcommands implementation that required to be tested.
- update to `go.mod` and `go.sum`


cmd/internal/tlogs.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func init() {
+	utilCmd.AddCommand(tailLogsCmd)
+}
+
+// WaitForFilesToAppear use a ticker to wait until files appear
+// under the specified dir, it will timeout after some time.
+func WaitForFilesToAppear(monitorDir string) ([]string, error) {
+	timeOut := time.After(120 * time.Second)
```
edwardstudy: maybe we could parameterize/add flags for these vars

cmd/internal/tlogs.go
```diff
@@ -0,0 +1,228 @@
...
+// - added italics font, for the file name
+func PrintOutput(messages chan StdOutMsg, items int) {
+	var colors []colorful.Color
+	colors = bunt.RandomTerminalFriendlyColors(items)
+
+	for msg := range messages {
+		humanReadableTime := fmt.Sprintf("%04d/%02d/%02d %02d:%02d:%02d", msg.Timestamp.Year(), msg.Timestamp.Month(), msg.Timestamp.Day(), msg.Timestamp.Hour(), msg.Timestamp.Minute(), msg.Timestamp.Second())
+
+		fmt.Printf("%s, %s, %s\n",
+			bunt.Style(humanReadableTime, bunt.Foreground(colors[msg.ID])),
```
edwardstudy: I am afraid that these color signs only were shown as plain text and it confuses users if we use this command in the sidebar.

qu1queee: will not, afaik.

PR #457, opened 2019-07-25
Update on implicit variable changes

edwardstudy:
[#167315929](https://www.pivotaltracker.com/story/show/167315929)

- Added `implicitVariables` field to BoshDeployment CRD and handled update on implicit variables
- Consolidated `manifest`, `ops` and `implicitVariables` type as `ResourceReference`

**Notice**:
It's a imcompatible change about BoshDeployment CRD that `spec.ops.ref`->`spec.ops.name`, `spec.manifest.ref`-> `spec.manifest.name`


edwardstudy:
Found wrong test assertions.

PR #459, opened 2019-07-25
Add reconciliation window support to errand extended jobs

manno:
After a successful reconcile, the timestamp is stored in an annotation
of the extended job. Further attempts to reconcile that ejob within
`MeltdownDuration` seconds will be delayed by `MeltdownRequeueAfter`
seconds.

Skipping ejobs due to missing inputs also requeued within micro seconds,
a config value was added and requeues now default to 1s.

[#166767434](https://www.pivotaltracker.com/story/show/166767434)


pkg/kube/controllers/extendedjob/errand_reconciler.go
```diff
@@ -70,6 +72,11 @@ func (r *ErrandReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 		return result, err
 	}
 
+	if meltdown.InWindow(time.Now(), r.config.MeltdownDuration, eJob.ObjectMeta.Annotations) {
+		ctxlog.WithEvent(eJob, "Meltdown").Debugf(ctx, "Resource '%s' is in meltdown, delaying reconciles for %s", eJob.Name, r.config.MeltdownDuration)
+		return reconcile.Result{Requeue: true, RequeueAfter: r.config.MeltdownRequeueAfter}, nil
```
edwardstudy: We could only set `RequeueAfter > 0` to requeue the request: https://github.com/cloudfoundry-incubator/cf-operator-vendor/blob/e337a0cdf3c31f7053be70212b3059d078ed8fa1/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L221

pkg/kube/controllers/extendedjob/errand_reconciler.go
```diff
@@ -100,6 +110,13 @@ func (r *ErrandReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 			err = ctxlog.WithEvent(eJob, "UpdateError").Errorf(ctx, "Failed to traverse to 'trigger.strategy=done' on job '%s': %s", eJob.Name, err)
 			return reconcile.Result{Requeue: false}, err
 		}
+	} else {
+		// several updates to one resource result in 'object has been modified' and the error triggers a second reconcile even though Requeue was false!?
```
edwardstudy: Yes: https://github.com/cloudfoundry-incubator/cf-operator-vendor/blob/e337a0cdf3c31f7053be70212b3059d078ed8fa1/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L215

Only return empty error if you want to skip this reconciliation.

manno:
This depends on #468

deploy/helm/cf-operator/templates/fissile_v1alpha1_extendedsecret_crd.yaml
```diff
@@ -18,4 +18,6 @@ spec:
         - esecs
   scope: Namespaced
   version: v1alpha1
+  subresources:
```
edwardstudy: CRD in code should be synced: https://github.com/cloudfoundry-incubator/cf-operator/blob/245486dcde519f54ca4a27ec92c1ee6443eb1247/pkg/kube/util/crd/crd.go#L18-L38

pkg/kube/util/meltdown/meltdown.go
```diff
@@ -0,0 +1,54 @@
...
+	return now.Before(windowEnd)
+}
+
+// NewWindow returns a new time window starting at lastReconcile, ending after duration
+func NewWindow(duration time.Duration, lastReconcile *metav1.Time) Window {
+	if lastReconcile != nil {
+		start := lastReconcile.Time
+		return Window{Start: start, Duration: duration}
+	}
+	return Window{}
```
edwardstudy: Is it we expect that return empty window, even nil Duration? do we have any case like that?

deploy/helm/cf-operator/templates/fissile_v1alpha1_extendedsecret_crd.yaml
manno: We'll move it to docs soon enough ;)

I think it's in code already: https://github.com/cloudfoundry-incubator/cf-operator/pull/459/files#diff-0c1a28ec9d4ea775ba31f02b722af57dR31

pkg/kube/util/meltdown/meltdown.go
manno: If you have no last reconcile timestamp this will return the year 0 and you will always be out of the duration.
Of course, this will break if your window duration is a few thousand years, but otherwise we're fine.

PR #465, opened 2019-07-29
Override EJobs' versioned output

edwardstudy:
[#167531317](https://www.pivotaltracker.com/story/show/167531317)

For a preview, I didn't add some unit test cases about new funcs.


edwardstudy:
Closed because we can endure this for secret immutability and concise.

PR #467, opened 2019-07-30
Allow arbitrary environment variables to be set for a POD

edwardstudy:
[#167316563](https://www.pivotaltracker.com/story/show/167316563)

- If we have same env key in `bpm.processes.env`, arbitrary env variables will override those.

```
IBMMasterMBP:~ bjxzi$ k exec -it -n scf scf-dev-log-api-v1-0 -c loggregator-trafficcontroller-loggregator-trafficcontroller -- bash
/:/# env | grep TRAFFIC_CONTROLLER_IP
TRAFFIC_CONTROLLER_IP=172.30.23.17
```

### DIFF
--- a/pkg/bosh/converter/container_factory.go
+++ b/pkg/bosh/converter/container_factory.go
@@ -199,6 +199,7 @@ func (c *ContainerFactory) JobsToContainers(
 				process,
 				processVolumeMounts,
 				job.Properties.Quarks.Run.HealthCheck,
+				job.Properties.Quarks.Envs,
 			)
 
 			containers = append(containers, *container.DeepCopy())
@@ -414,6 +415,7 @@ func bpmProcessContainer(
 	process bpm.Process,
 	volumeMounts []corev1.VolumeMount,
 	healthchecks map[string]bdm.HealthCheck,
+	arbitraryEnvs []corev1.EnvVar,
 ) corev1.Container {
 	name := names.Sanitize(fmt.Sprintf("%s-%s", jobName, processName))
 	container := corev1.Container{
@@ -475,9 +477,7 @@ func bpmProcessContainer(
 		},
 	}
 
-	for name, value := range process.Env {
-		container.Env = append(container.Env, corev1.EnvVar{Name: name, Value: value})
-	}
+	container.Env = generateEnv(process.Env, arbitraryEnvs)
 
 	for name, hc := range healthchecks {
 		if name == process.Name {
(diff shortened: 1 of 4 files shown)


pkg/bosh/converter/container_factory.go
```diff
@@ -525,3 +532,15 @@ func flattenContainers(containers ...interface{}) []corev1.Container {
 	}
 	return result
 }
+
+func updateEnvs(envs []corev1.EnvVar, env corev1.EnvVar) []corev1.EnvVar {
+	for i, envVar := range envs {
+		if envVar.Name == env.Name {
+			envs[i] = env
```
f0rmiga: I dislike a lot this parameter being mutated. Ideally, we should generate the `container.Env` and treat it as immutable.

You will want something like (not tested - wrote directly here -, and can probably be optimized):
```go
func overrideEnvs(envs, overrides []corev1.EnvVar) []corev1.EnvVar {
  result := make([]corev1.EnvVar, 0, len(envs))
  overridesMap := make(map[string]corev1.EnvVar)
  for _, override := range overrides {
    overridesMap[override.Name] = override
  }
  for _, env := range envs {
    if override, ok := overridesMap[env.Name]; ok {
      result = append(result, override)
    } else {
      result = append(result, env)
    }
  }
}
```

The output of this function is then used as the `container.Env` above.

pkg/bosh/manifest/containerization.go
```diff
@@ -10,14 +10,15 @@ import (
...
 	Consumes         map[string]JobLink `json:"consumes"`
-	Instances        []JobInstance      `json:"instances"`
-	Release          string             `json:"release"`
-	BPM              *bpm.Config        `json:"bpm,omitempty" yaml:"bpm,omitempty"`
-	Ports            []Port             `json:"ports"`
-	Run              RunConfig          `json:"run"`
-	PreRenderScripts []string           `json:"pre_render_scripts" yaml:"pre_render_scripts"`
-	Debug            bool               `json:"debug" yaml:"debug"`
-	IsAddon          bool               `json:"is_addon" yaml:"is_addon"`
+	Instances        []JobInstance   `json:"instances"`
```
f0rmiga: Doesn't seem the formatter ran here.

pkg/bosh/converter/container_factory.go
```diff
@@ -479,6 +481,11 @@ func bpmProcessContainer(
 		container.Env = append(container.Env, corev1.EnvVar{Name: name, Value: value})
 	}
 
+	// arbitrary envs should override envs
```
f0rmiga: [nit] Comments should be capitalized and punctuated.

pkg/bosh/manifest/containerization_test.go
```diff
@@ -44,4 +45,22 @@ var _ = Describe("BOSHContainerization", func() {
...
+		Expect(envs[0]).To(Equal(corev1.EnvVar{
+			Name: "TRAFFIC_CONTROLLER_IP",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					APIVersion: "v1",
+					FieldPath:  "status.podIP",
+				},
+			},
+		}))
+		//Expect(healthchecks["doppler"].ReadinessProbe).To(BeNil())
```
f0rmiga: Commented code should be removed.

pkg/bosh/converter/container_factory.go
```diff
@@ -479,6 +481,11 @@ func bpmProcessContainer(
 		container.Env = append(container.Env, corev1.EnvVar{Name: name, Value: value})
 	}
 
+	// Arbitrary envs should override container.Env.
+	if len(arbitraryEnvs) != 0 {
+		container.Env = overrideEnv(container.Env, arbitraryEnvs)
```
f0rmiga: In some sense, this is still mutating `container.Env`. Ideally, you wouldn't set `container.Env = <something>`, but rather initialize the `corev1.Container{}` above with the envs already. What you can do is to combine this with the logic of the `range process.Env` to be done before `container` is instantiated.

PR #472, opened 2019-08-01
Post-start mechanism

f0rmiga:
Added the post-start mechanism to mimic how BOSH handles it, and a few other minor fixes.


edwardstudy:
Hi, @f0rmiga. Please rebase from the master. Because we renamed `bosh_containerization` to `quarks`.  Thx.

container-run/pkg/containerrun/cmd.go
```diff
@@ -0,0 +1,62 @@
+package containerrun
```
manno: I would have expected this to live below `cmd/containerrun`. I remember the cobra docs suggested a sub folder, if you intend to have multiple binaries. That's why it's `cmd/cf-operator/main.go`.

go.mod
```diff
@@ -18,36 +16,34 @@ require (
 	github.com/go-test/deep v1.0.3
 	github.com/gogo/protobuf v1.2.1 // indirect
 	github.com/golang/groupcache v0.0.0-20180924190550-6f2cf27854a4 // indirect
+	github.com/golang/mock v1.3.1
```
manno: We already have counterfeiter (and `bin/gen-fakes`), which is used in other bosh projects, too. 
Besides being more popular, does mockgen have any advantages in your opinion?

PR #477, opened 2019-08-05
Make staticcheck happy

aduffeck:


cmd/internal/root.go
```diff
@@ -14,16 +14,16 @@ import (
 	"go.uber.org/zap"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc" // from https://github.com/kubernetes/client-go/issues/345
+	crlog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
```
edwardstudy: what's the difference between `manager ` and `runtime`?

aduffeck: `runtime/signals` is deprecated, it's merely a wrapper around `manager/signals` at this point. See https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/runtime/signals/signal.go#L21

PR #479, opened 2019-08-06
Fix Resolver path in make target

rohitsakala:


rohitsakala:
> Could you please sync these fake file names or remove them in code:
> https://github.com/cloudfoundry-incubator/cf-operator/blob/7cbbaf9a9af6134fde769f02b1da22811f38b7a7/pkg/bosh/converter/resolver.go#L23

removed them.

PR #480, opened 2019-08-06
quarks.pre_render_scripts should be separate

qu1queee:
[#167476372](https://www.pivotaltracker.com/story/show/167476372)
This modify the pre_render_scripts struct definition, in favour of
a new PreRender struct with three fields for different types of scripts:
- bpm
- ig_resolver
- jobs

This would ensure that scripts can run separately, and only when define.

### DIFF
--- a/pkg/bosh/manifest/cmd_render_job_tmpls.go
+++ b/pkg/bosh/manifest/cmd_render_job_tmpls.go
@@ -13,6 +13,12 @@ import (
 	btg "github.com/viovanov/bosh-template-go"
 )
 
+const (
+	typeBPM        = "bpm"
+	typeIGResolver = "ig_resolver"
+	typeJobs       = "jobs"
+)
+
 // RenderJobTemplates will render templates for all jobs of the instance group
 // https://bosh.io/docs/create-release/#job-specs
 // boshManifest is a resolved manifest for a single instance group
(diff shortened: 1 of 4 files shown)


pkg/bosh/converter/container_factory.go
```diff
@@ -377,13 +377,7 @@ func boshPreStartInitContainer(
...
 }
 
-func bpmPreStartInitContainer(
-	process bpm.Process,
-	jobImage string,
-	volumeMounts []corev1.VolumeMount,
-	debug bool,
-) corev1.Container {
-
+func bpmPreStartInitContainer(process bpm.Process, jobImage string, volumeMounts []corev1.VolumeMount, debug bool) corev1.Container {
```
f0rmiga: Please, avoid those long lines. It's much easier to read a multi-line parameter list. And also much easier to review parameters added/removed.

pkg/bosh/converter/container_factory.go
```diff
@@ -408,15 +402,7 @@ func bpmPreStartInitContainer(
...
-func bpmProcessContainer(
-	jobName string,
-	processName string,
-	jobImage string,
-	process bpm.Process,
-	volumeMounts []corev1.VolumeMount,
-	healthchecks map[string]bdm.HealthCheck,
-	arbitraryEnvs []corev1.EnvVar,
-) corev1.Container {
+func bpmProcessContainer(jobName string, processName string, jobImage string, process bpm.Process, volumeMounts []corev1.VolumeMount, healthchecks map[string]bdm.HealthCheck, arbitraryEnvs []corev1.EnvVar) corev1.Container {
```
f0rmiga: Same here.

pkg/bosh/manifest/cmd_render_job_tmpls.go
```diff
@@ -113,59 +119,79 @@ func RenderJobTemplates(
 	return nil
 }
 
-func runPreRenderScripts(instanceGroup *InstanceGroup, silent bool) error {
-	for _, job := range instanceGroup.Jobs {
-		for idx, script := range job.Properties.Quarks.PreRenderScripts {
-			createErr := func(err error) error {
-				return errors.Wrapf(err, "failed to run pre-render script %d for job %s for instance group %s", idx, job.Name, instanceGroup.Name)
-			}
+func runRenderScript(jobName string, typeScript string, scripts []string, igName string, silent bool) error {
```
f0rmiga: I'd appreciate if this parameter list could become multi-line as well.

f0rmiga: `scriptType` sounds more natural than `typeScript` (which is also a programming language :laughing:).

pkg/bosh/manifest/containerization.go
```diff
@@ -58,3 +58,11 @@ type HealthCheck struct {
 type RunConfig struct {
 	HealthCheck map[string]HealthCheck `json:"healthcheck" yaml:"healthcheck"`
 }
+
+// PreRender describes the different types of scripts
+// that can be run inside a job
+type PreRender struct {
```
f0rmiga: Let's name this `PreRenderScripts` (it's fine it being plural here).

pkg/bosh/manifest/containerization.go
```diff
@@ -58,3 +58,11 @@ type HealthCheck struct {
...
+	BpmScripts        []string `json:"bpm" yaml:"bpm"`
```
f0rmiga: Let's match the field names with the json ones. E.g.
```
BPM []string `json:"bpm" yaml:"bpm"`
```

PR #482, opened 2019-08-06
Add staticcheck make target

qu1queee:
This is needed so that we can have a convenience way of verifying
we are not breaking the staticcheck check in the CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ check-scripts:
 	bin/check-scripts
 
 staticcheck:
-    bin/staticcheck
+	bin/staticcheck
 
 ############ BUILD TARGETS ############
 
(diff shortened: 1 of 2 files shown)


bin/staticcheck
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+if ! hash staticcheck2  2>/dev/null; then
```
aduffeck:
```suggestion
if ! hash staticcheck  2>/dev/null; then
```

Makefile
```diff
@@ -17,6 +17,9 @@ tools:
 check-scripts:
 	bin/check-scripts
 
+staticcheck:
+    bin/staticcheck
```
aduffeck: This needs to be a tab afaict.

PR #484, opened 2019-08-07
Support ExtendedSecret certs signed by the Kube API Server

edwardstudy:
[#167555390](https://www.pivotaltracker.com/story/show/167555390)

- Added two options to support CSR for ExtendedSecret:
```
SignerType       SignerType        `json:"signerType,omitempty"`
Usages           []certv1.KeyUsage `json:"usages,omitempty"`
```

- Added CSR controller to watch CSR and approve it and then persists certificate signed by Kube API Server

Verification steps: 
```
$ k apply -f ./docs/examples/extended-secret/certificate.yaml

$ k get secret
NAME                                   TYPE                                  DATA   AGE
...
nats-deployment.var-nats-cert    # This cert signed by API Server
```

### DIFF
--- a/docs/controllers/extendedsecret.md
+++ b/docs/controllers/extendedsecret.md
@@ -6,6 +6,7 @@
   - [Features](#features)
     - [Generated](#generated)
     - [Policies](#policies)
+    - [Auto-approving Certificates](#auto-approving-certificates)
   - [`ExtendedSecret` Examples](#extendedsecret-examples)
 
 ## Description
@@ -31,6 +32,24 @@ A pluggable implementation for generating certificates and passwords.
 ### Policies
 
 The developer can specify policies for rotation (e.g. automatic or not) and how secrets are created (e.g. password complexity, certificate expiration date, etc.).
+
+### Auto-approving Certificates
+
+A certificate `ExtendedSecret` can be signed by the Kube API Server. The ExtendedSecret Controller is be responsible for generating certificate signing request and creating CertificateSigningRequest:
+
+```yaml
+apiVersion: certificates.k8s.io/v1beta1
+kind: CertificateSigningRequest
+metadata:
+  name: generate-certificate
+spec:
+  request: ((encoded-cert-signing-request))
+  usages:
+  - digital signature
+  - key encipherment
+```
+
+The CertificateSigningRequest controller watches for `CertificateSigningRequest` and approves ExtendedSecret-owned CSRs and persists the generated certificate.
 
 ## `ExtendedSecret` Examples
 
--- a/docs/controllers/extendedsecret.md
+++ b/docs/controllers/extendedsecret.md
@@ -15,9 +15,14 @@
 
 `ExtendedSecret` supports generating the following:
 
-- certificates
-- passwords
-- rsa keys
+| Secret Type                     | spec.type     | certificate.signerType | certificate.isCA    |
+| ------------------------------- | ------------- | ---------------------- | ------------------- |
+| `passwords`                     | `password`    | not set                | not set             |
+| `rsa keys`                      | `rsa`         | not set                | not set             |
+| `ssh keys`                      | `ssh`         | not set                | not set             |
+| `self-signed root certificates` | `certificate` | `local`                | `true`              |
+| `self-signed certificates`      | `certificate` | `local`                | `false`             |
+| `cluster-signed certificates`   | `certificate` | `cluster`              | `false`             |
 
 > **Note:**
 >
@@ -35,7 +40,7 @@ The developer can specify policies for rotation (e.g. automatic or not) and how 
 
 ### Auto-approving Certificates
 
-A certificate `ExtendedSecret` can be signed by the Kube API Server. The ExtendedSecret Controller is be responsible for generating certificate signing request and creating CertificateSigningRequest:
+A certificate `ExtendedSecret` can be signed by the Kube API Server. The ExtendedSecret Controller is be responsible for generating certificate signing request and approving the request:
 
 ```yaml
 apiVersion: certificates.k8s.io/v1beta1
--- a/docs/examples/extended-secret/certificate.yaml
+++ b/docs/examples/extended-secret/certificate.yaml
@@ -1,94 +1,4 @@
 ---
-apiVersion: v1
-kind: Secret
-metadata:
-  name: cert-ca
-type: Opaque
-stringData:
-  certificate: |-
-    -----BEGIN CERTIFICATE-----
-    MIIE9jCCAt6gAwIBAgIUEtrf/IIROoFcpubnPMMaNc00cwIwDQYJKoZIhvcNAQEN
-    BQAwEzERMA8GA1UEAxMIcm91dGVyQ0EwHhcNMTkwODA3MDQxNjAwWhcNMjAwODA2
-    MDQxNjAwWjATMREwDwYDVQQDEwhyb3V0ZXJDQTCCAiIwDQYJKoZIhvcNAQEBBQAD
-    ggIPADCCAgoCggIBAMs548l/BVFJivKZzXyv4koxHNI7suvyod6TroTE/xYDs13w
-    wk0oI7jrQ2/bz94ospcAoFDEWDA3PLzaueHJ4DG942pwp5nrq0bWrskf3i6dYqfK
-    defZ5huB769YKG+w1da4jkesohwIHtBU/sOu+lSXUpbvCnEAUjKhKbKee/8vUQWI
-    s1s2mLqsxC3jiBFXaeoo6m/uPZYYwsWgTG3OMOjskpOIGsIp3HU+F8NeNoBcuTFu
-    k2+GVBqHAiif1Tcoc4YXZst0v16sQL+WWmbb3d/mT4raqeWu/ni1X5ZOrNGn9wzN
-    EvUTDX4RmWAvj0FscYOvcxOh+ZacjVSrfjCksZGem+R+1XWR6qQHZQ77OGtMPe+6
-    9LBtxPpIPRF9+79iQwDiqFm1abG0os8A3HM0465qTU3UkBy8UuBCEM2BTEr9518/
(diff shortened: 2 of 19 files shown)


pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
```diff
@@ -0,0 +1,209 @@
...
+
+		err = r.deleteSecret(ctx, privatekeySecret)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{}, nil
+	}
+
+	err = r.approveRequest(ctx, instance.Name)
```
aduffeck: I think this should better go into an `else` branch.

pkg/kube/util/names/names.go
```diff
@@ -180,6 +180,11 @@ func OrdinalFromPodName(name string) int {
 	return podOrdinal
 }
 
+// PrivateKeySecretName returns a Secret name for a given ExtendedSecret
+func PrivateKeySecretName(extendedSecretName string) string {
```
aduffeck: Could we make this name a bit more precise, i.e. ` CsrPrivateKeySecretName`?

pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
```diff
@@ -0,0 +1,209 @@
...
+			ctxlog.Info(ctx, "Skip reconcile: certificateSigningRequest not found")
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		ctxlog.Info(ctx, "Error reading the object")
+		return reconcile.Result{}, errors.Wrap(err, "Error reading certificateSigningRequest")
+	}
+
+	if len(instance.Status.Certificate) != 0 {
+		annotations := instance.GetAnnotations()
```
aduffeck: I'm just wondering, why do we need these annotations? Couldn't we get the secret name and namespace from the owner object (ExtendedSecret)?

pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		namespace, ok := annotations[esv1.AnnotationExSecretNamespace]
+		if !ok {
+			ctxlog.WithEvent(instance, "NotFoundError").Errorf(ctx, "failed to lookup exSecret namespace from certificateSigningRequest '%s' annotations", instance.Name)
+			return reconcile.Result{}, nil
+		}
+
+		privatekeySecret, err := r.getSecret(ctx, namespace, names.PrivateKeySecretName(instance.Name))
+		if err != nil {
+			return reconcile.Result{}, err
```
aduffeck: How about adding a log message here?

pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
aduffeck: And couldn't we pass the object directly instead of its name to get rid of another get request?

pkg/kube/controllers/extendedsecret/extendedsecret_reconciler.go
```diff
@@ -124,6 +127,7 @@ func (r *ReconcileExtendedSecret) Reconcile(request reconcile.Request) (reconcil
 func (r *ReconcileExtendedSecret) updateExSecret(ctx context.Context, instance *esv1.ExtendedSecret) error {
 	obj := instance.DeepCopy()
 	op, err := controllerutil.CreateOrUpdate(ctx, r.client, obj, func() error {
+		obj.Spec = instance.Spec
```
aduffeck: Why is this change needed?

pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
edwardstudy: The truth is that:
1. CSR is no-namespaced
2. the owner and owned object are both namespace-scoped and in the same namespace

So getting the owner's namespace from CSR could not get right namespace. 
My idea is to set the owner's namespace to annotations.

pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
edwardstudy: Oh, it's helpful for reducing update conflicts. 
And `r.client` could not update subresource `approval`. we have to use `client-go.client` to update CSR approval directly: 
https://github.com/kubernetes-sigs/controller-runtime/issues/172

pkg/kube/controllers/extendedsecret/extendedsecret_reconciler.go
edwardstudy: It's helpful to update default `spec.request.signerType` if it's certificate.

docs/examples/extended-secret/certificate.yaml
```diff
@@ -0,0 +1,112 @@
...
+        Name: cert-ca
+      CARef:
+        Key: certificate
+        Name: cert-ca
+      alternativeNames:
+        - foo.com
+        - '*.foo.com'
+      commonName: routerSSL
+      isCA: false
+      signerType: external
```
viovanov: This looks like it's not using the API server to sign

docs/controllers/extendedsecret.md
```diff
@@ -32,6 +33,24 @@ A pluggable implementation for generating certificates and passwords.
 
 The developer can specify policies for rotation (e.g. automatic or not) and how secrets are created (e.g. password complexity, certificate expiration date, etc.).
 
+### Auto-approving Certificates
+
+A certificate `ExtendedSecret` can be signed by the Kube API Server. The ExtendedSecret Controller is be responsible for generating certificate signing request and creating CertificateSigningRequest:
```
viovanov: For generating ... And _approving_

pkg/credsgen/in_memory_generator/certificate.go
```diff
@@ -38,6 +38,27 @@ func (g InMemoryGenerator) GenerateCertificate(name string, request credsgen.Cer
...
+// GenerateCertificateSigningRequest Generates a certificate signing request and private key
+func (g InMemoryGenerator) GenerateCertificateSigningRequest(request credsgen.CertificateGenerationRequest) ([]byte, []byte, error) {
+	cfssllog.Level = cfssllog.LevelWarning
+
+	var csReq, privateKey []byte
+
+	// Generate certificate request
+	certReq := &csr.CertificateRequest{KeyRequest: &csr.BasicKeyRequest{A: g.Algorithm, S: g.Bits}}
+
+	certReq.Hosts = append(certReq.Hosts, request.CommonName)
```
viovanov: Do you know if it works with IPs?

pkg/kube/apis/extendedsecret/v1alpha1/types.go
```diff
@@ -12,20 +13,33 @@ import (
...
+	Certificate SecretType = "certificate"
+	SSHKey      SecretType = "ssh"
+	RSAKey      SecretType = "rsa"
+)
+
+// SignerType defines the type of the certificate signer
+type SignerType = string
+
+// Valid values for signer types
+const (
```
viovanov: What do these mean?

viovanov: Can you please document?

docs/examples/extended-secret/certificate.yaml
edwardstudy: Removed CA infos. Sorry for misleading example.

pkg/kube/apis/extendedsecret/v1alpha1/types.go
edwardstudy: Done

pkg/credsgen/in_memory_generator/certificate.go
edwardstudy: It should be. thinking a case about it.

pkg/kube/controllers/extendedsecret/certificatesigningrequest_reconciler.go
edwardstudy: done

pkg/kube/controllers/extendedsecret/extendedsecret_reconciler.go
edwardstudy: refactored CreateOrUpdate mutateFn

edwardstudy:
operator can consume following CR and create cert signed by API server
```
---
apiVersion: fissile.cloudfoundry.org/v1alpha1
kind: ExtendedSecret
metadata:
  name: generate-certificate
spec:
  request:
    certificate:
      alternativeNames:
        - foo.com
        - '*.foo.com'
      commonName: routerSSL
      signerType: cluster
      usages: ["server auth", "client auth"]
  secretName: gen-certificate
  type: certificate
```

e2e/kube/examples_test.go
```diff
@@ -170,33 +170,33 @@ var _ = Describe("Examples", func() {
...
+					Expect(err).ToNot(HaveOccurred(), "error restarting cf-operator")
 
 					By("Checking for pods not created")
 					err = kubectlHelper.Wait(namespace, "ready", "pod/nats-deployment-nats-v2-0", 10*time.Second)
 					Expect(err).To(HaveOccurred(), "error unexpected version of instance group is created")
 
 					By("Checking for secrets not created")
 					exist, err := kubectlHelper.SecretExists(namespace, "nats-deployment.bpm.nats-v2")
-					Expect(err).To(HaveOccurred())
+					Expect(err).ToNot(HaveOccurred(), "error getting secret/nats-deployment.bpm.nats-v2")
```
aduffeck: Why was the expectation negated here and in the cases below?

edwardstudy: This case is that we don't want new versions to be created when the operator restart. I changed `kubectlHelper.SecretExists` that doesn't return an error when object not found because I want to use it in the WaitForSecret func that didn't break poll when object not found.

aduffeck: Got it.

PR #485, opened 2019-08-08
Fix flaky exjob test

rohitsakala:
[#167647818](https://www.pivotaltracker.com/n/projects/2192232/stories/167647818)

Return the earliest created jobpod when multiple jobpods exists
as the kubernetes job creates a new pod every time a previously
created pod by it fails.

Also, adding custom backofflimit to 2 as the default 6 exceeds the integration timeout. 
Mentioned about this change in the story. [0]

[0] https://www.pivotaltracker.com/n/projects/2192232/stories/167866215


pkg/kube/controllers/extendedjob/job_reconciler.go
```diff
@@ -154,7 +154,25 @@ func (r *ReconcileJob) jobPod(ctx context.Context, name string, namespace string
...
 		return nil, errors.Errorf("Job %s does not own any pods?", name)
 	}
-	return &list.Items[0], nil
+
+	// If there is only one jobpod, then return index 0 pod
+	if len(list.Items) == 1 {
+		return &list.Items[0], nil
+	} else {
+		// If there are more than one jobpods, then return the earliest
+		// created jobpod. There will be multiple jobpods when jobpods
```
aduffeck: Why would you want the earliest created pod returned? I'd have thought that you'd be more interested in the latest, possibly succeeded one.

Do you mind adding a test for this change?

rohitsakala: Sure, I will add a test for it. 

Yes, it makes sense to check for the latest change. But by reading the code of the `job_controller`, I realized we only care about the first job pod. In this line [0], the `o.Status.Failed` count is only 1, but when there is a failed test, the count will be 2. When the count is 2, we don't reconcile as per the condition.

Let me know if we should change this ?

[0] https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/controllers/extendedjob/job_controller.go#L57

pkg/kube/controllers/extendedjob/job_controller.go
```diff
@@ -54,7 +54,7 @@ func AddJob(ctx context.Context, config *config.Config, mgr manager.Manager) err
 				return false
 			}
 
-			shouldProcessEvent := o.Status.Succeeded == 1 || o.Status.Failed == 1
+			shouldProcessEvent := o.Status.Succeeded == 1 || o.Status.Failed == (*o.Spec.BackoffLimit+1)
```
aduffeck: Not sure if this can actually happen, but maybe the `BackoffLimit` could be decreased by the user and thus the `Failed` count becomes greather than `BackoffLimit+1`?

```suggestion
			shouldProcessEvent := o.Status.Succeeded == 1 || o.Status.Failed >= (*o.Spec.BackoffLimit+1)
```

pkg/kube/controllers/extendedjob/job_creator.go
```diff
@@ -103,13 +103,19 @@ func (j jobCreatorImpl) Create(ctx context.Context, eJob ejv1.ExtendedJob, podNa
 	if err != nil {
 		return false, errors.Wrapf(err, "could not generate job name for eJob '%s'", eJob.Name)
 	}
+
+	backoffLimit := int32(2)
```
aduffeck: Oh, I thought we decided to not change the defaults during the call? Or did I miss anything?

pkg/kube/controllers/extendedjob/job_reconciler_test.go
```diff
@@ -175,23 +179,39 @@ var _ = Describe("ReconcileExtendedJob", func() {
...
+							Items: []corev1.Pod{*pod1, *pod2},
+						}
+						list.DeepCopyInto(object)
+					}
+					return nil
+				})
+
 				result, err := reconciler.Reconcile(request)
 				Expect(err).ToNot(HaveOccurred())
+				Expect(logs.FilterMessageSnippet("Considering job pod foo-job-latest for persisting output").Len()).To(Equal(1))
```
aduffeck: I think it would be good to not use a log message for verifying that the latest pod is being used for persistance but rather something lower level. Log messages tend to change and can lie ;)

pkg/kube/controllers/extendedjob/job_creator.go
rohitsakala: Added in the story -  https://www.pivotaltracker.com/n/projects/2192232/stories/167866215

pkg/kube/controllers/extendedjob/job_reconciler_test.go
rohitsakala: Changed it. Thanks for the solution.

pkg/kube/controllers/extendedjob/job_controller.go
```diff
@@ -54,7 +54,7 @@ func AddJob(ctx context.Context, config *config.Config, mgr manager.Manager) err
 				return false
 			}
 
-			shouldProcessEvent := o.Status.Succeeded == 1 || o.Status.Failed == 1
+			shouldProcessEvent := o.Status.Succeeded == 1 || o.Status.Failed >= (*o.Spec.BackoffLimit+1)
```
edwardstudy: why not `o.Status.Failed > (*o.Spec.BackoffLimit)`?

pkg/kube/controllers/extendedjob/job_creator.go
edwardstudy: Use `Int32()` in `pkg/kube/util/pointer.go` is better.

rohitsakala: Done.

pkg/kube/controllers/extendedjob/job_controller.go
rohitsakala: Done.

PR #486, opened 2019-08-08
Add DisableLogSidecar field into AgentSettings struct

qu1queee:
This should allow users to disable the default sidecar
running in all pods for logging purposes.

If you will like to disable it, you need to define the following under the `instance_group` item:
```
      env:
        bosh:
          agent: 
            settings:
              disable_log_sidecar: true
```

### DIFF
--- a/docs/examples/bosh-deployment/boshdeployment-with-custom-variable-disable-sidecar.yaml
+++ b/docs/examples/bosh-deployment/boshdeployment-with-custom-variable-disable-sidecar.yaml
@@ -1,0 +1,64 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: nats-deployment.var-custom-password
+type: Opaque
+stringData:
+  password: custom_password
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nats-manifest
+data:
+  manifest: |
+    ---
+    name: nats-manifest
+    releases:
+    - name: nats
+      version: "26"
+      url: docker.io/cfcontainerization
+      stemcell:
+        os: opensuse-42.3
+        version: 30.g9c91e77-30.80-7.0.0_257.gb97ced55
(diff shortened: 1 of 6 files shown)


pkg/bosh/converter/bpm_resources.go
```diff
@@ -109,7 +109,9 @@ func (kc *KubeConverter) BPMResources(manifestName string, version string, insta
 		Disks: allDisks,
 	}
 
-	cfac := NewContainerFactory(manifestName, instanceGroup.Name, version, releaseImageProvider, bpmConfigs)
+	disableLogSideCar := instanceGroup.Env.AgentEnvBoshConfig.Agent.Settings.DisableLogSidecar
+
+	cfac := NewContainerFactory(manifestName, instanceGroup.Name, version, disableLogSideCar, releaseImageProvider, bpmConfigs)
```
edwardstudy: why not use `instanceGroup.Env.AgentEnvBoshConfig.Agent.Settings.DisableLogSidecar` directly?

docs/examples/bosh-deployment/boshdeployment-with-custom-variable-disable-sidecar.yaml
```diff
@@ -0,0 +1,64 @@
+---
```
edwardstudy: If you write an example file, could you add an e2e case for it?
https://ci.flintstone.cf.cloud.ibm.com/builds/92928#L5d27a265:531

qu1queee: yes, i missed it, working on it

pkg/bosh/converter/bpm_resources.go
qu1queee: yes, let me change it

PR #488, opened 2019-08-12
Make more precise changes for `CreateOrUpdates`

edwardstudy:
[#167523284](https://www.pivotaltracker.com/story/show/167523284)


edwardstudy:
Hi, @viovanov. I checked `v0.2.0-beta.5` and we used `v0.2.0-beta.4`. they are suitable. 
My idea is that only updating `labels`, `annotations`, and `spec` and those private functions will return mutateFn after I read test cases of `CreateOrUpdate`:

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.2.0-beta.5/pkg/controller/controllerutil/controllerutil_test.go#L276-L281

PR #489, opened 2019-08-13
Don't ever update BOSHDeployment

viovanov:


aduffeck:
I can confirm that this fixes the `context deadline exceeded` issues we were seeing.

PR #491, opened 2019-08-14
Fix test cases and revert DesiredManifest func signature 

edwardstudy:
It's a hotfix


edwardstudy:
It seemed that e2e cases failed also, fixing...👷

PR #494, opened 2019-08-16
Remove unnecessary double-filtering in predicates

aduffeck:


pkg/kube/controllers/boshdeployment/deployment_controller.go
```diff
@@ -49,30 +49,14 @@ func AddDeployment(ctx context.Context, config *config.Config, mgr manager.Manag
...
-		CreateFunc: func(e event.CreateEvent) bool {
-			configMap := e.Object.(*corev1.ConfigMap)
-
-			reconciles, err := reference.GetReconciles(ctx, mgr.GetClient(), reference.ReconcileForBOSHDeployment, configMap)
-			if err != nil {
-				ctxlog.Errorf(ctx, "Failed to calculate reconciles for configMap '%s': %v", configMap.Name, err)
-			}
-
-			// The ConfigMap should reference at least one BOSHDeployment in order for us to consider it
-			return len(reconciles) > 0
```
edwardstudy: what's the benefit that we process all configMap creation events?

aduffeck: We don't process them. The irrelevant ones are filtered in the map func at https://github.com/cloudfoundry-incubator/cf-operator/pull/494/files#diff-cc3ab76cd66ceda6b7f3115e24fd6d1bR70

Until now we applied the same filtering logic twice, once in the predicate and then again in the map function. We first got the list of required reconciles in the predicate and only continued if there was at least on reconcile needed. Then in the map func we got that same list again and triggered the according reconciles. So if that list is empty no reconciles are triggered.

Thus just getting the list in the map func has exactly the same effect, the predicate doesn't add any filtering and should be removed. Esp. since calculating the list of reconciles can be an expensive operation.

PR #496, opened 2019-08-19
Add missing docs

qu1queee:
[#167487168](https://www.pivotaltracker.com/story/show/167487168)

For the pre_render_scripts feature

### DIFF
--- a/docs/from_bosh_to_kube.md
+++ b/docs/from_bosh_to_kube.md
@@ -25,6 +25,7 @@
     - [Readiness and Liveness Probes](#Readiness-and-Liveness-Probes)
     - [Persistent Disks](#Persistent-Disks)
     - [Manual ("implicit") variables](#Manual-%22implicit%22-variables)
+    - [Pre_render_scripts](#Pre_render_scripts)
   - [Flow](#Flow)
   - [Naming Conventions](#Naming-Conventions)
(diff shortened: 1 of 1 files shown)


docs/from_bosh_to_kube.md
```diff
@@ -541,6 +542,37 @@ stringData:
   value: example.com
 ```
 
+### Pre_render_scripts
+
+Similar to what can be achieved in SCF v1, with the [patches](https://github.com/SUSE/scf/tree/develop/container-host-files/etc/scf/config/scripts/patches) scripts, the `cf-operator` is able to support this behaviour. Basically, it allows the user to execute a custom script during runtime of the job container for a specific `instance_group`. Because patching during runtime is always a great feature to have, for a variety of reasons, users can specify this via the `quarks.pre_render_scripts` key. 
+
+Keep it mind, that the script should belong to a type, to avoid running all scripts as a whole. Currently supported types are:
+
+- `quarks.pre_render_scripts.bpm`.
```
rohitsakala: Is there any difference between these three types? If so, do you think we should explain that?

qu1queee: From the code perspective not. From the syntax yes, you should define your script under the field that fits best for it.

PR #497, opened 2019-08-19
Extract all the different volume and volume mount structs

manno:
Even though the construction code looks identical and it's just structs, they do have a real world representation.
The new, separate functions make parameters explicit, document the volumes
purpose and hide construction details.

Maybe this can be improved upon, but I feel like it's a good first step to make the different volumes more explicit.

[#167507452](https://www.pivotaltracker.com/story/show/167507452)


pkg/bosh/converter/container_factory.go
```diff
@@ -272,19 +267,9 @@ func templateRenderingContainer(instanceGroupName string, secretName string) cor
...
-				MountPath: VolumeJobsDirMountPath,
-			},
-			{
-				Name:      generateVolumeName(secretName),
-				MountPath: fmt.Sprintf("/var/run/secrets/resolved-properties/%s", instanceGroupName),
-				ReadOnly:  true,
-			},
+			*renderingVolumeMount(),
+			*jobsDirVolumeMount(),
+			resolvedPropertiesVolumeMount(secretName, instanceGroupName),
```
edwardstudy: Why only this return copy but others return pointer?

manno: Because they are also used in `volume.go` and there a pointer is expected. I didn't want to introduce temporary variables just to get the reference in volume.go.

But yes, this is one of the surprising things that turned up while doing this refactoring. Sometimes we use pointers, sometimes we don't. I don't think it's bad and has to be unified now, but definitely one of the parts where this PR could be improved upon.

PR #501, opened 2019-08-22
Add retry logic for trigger jobs creation

edwardstudy:
[#167452931](https://www.pivotaltracker.com/story/show/167452931)

- used ExponetialBackoff to create kube jobs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/daaku/go.zipexe v1.0.1 // indirect
 	github.com/dchest/uniuri v0.0.0-20160212164326-8902c56451e9
 	github.com/fsnotify/fsnotify v1.4.7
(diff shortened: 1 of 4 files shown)


manno:
Isn't it sufficient to re-queue the complete reconcile?

https://github.com/manno/quarks-ejob/blob/master/pkg/kube/controllers/extendedjob/errand_reconciler.go#L79-L88

edwardstudy:
@manno I think we can close it because we remove trigger jobs. The retry is working for trigger jobs:

```
			if _, err := r.jobCreator.Create(ctx, eJob, podName, string(pod.UID)); err != nil {
				ctxlog.WithEvent(&eJob, "CreateJob").Infof(ctx, "Failed to create job for '%s' via pod %s: %s", eJob.Name, podEvent, err)
				continue
			}
```

PR #502, opened 2019-08-22
Increase Reconcile ctx timeouts

qu1queee:
[#167968068](https://www.pivotaltracker.com/story/show/167968068)

For every call we do to the API for created/update/delete events, we pass a context with a deadline.

Increase the default timeout from 10s to 30s. While for SCF v3, it seems that during many of this api calls, the current times are very close to the deadline of the context. Leading to `deadline exceeded` errors.

### DIFF
--- a/pkg/kube/util/config/config.go
+++ b/pkg/kube/util/config/config.go
@@ -8,7 +8,7 @@ import (
 
 const (
 	// CtxTimeOut is the default context.Context timeout
-	CtxTimeOut = 10 * time.Second
+	CtxTimeOut = 30 * time.Second
 )
 
 // Config controls the behaviour of different controllers


qu1queee:
@edwardstudy an option of what type do you have in mind? you can set it on the `Config` struct, and pass it to the operator `NewManager()` func

edwardstudy:
> @edwardstudy an option of what type do you have in mind? you can set it on the `Config` struct, and pass it to the operator `NewManager()` func

This value is a default, could we set it over cf-operator option.

https://github.com/cloudfoundry-incubator/cf-operator/blob/c6b81d0a32e7c9fae00e4b0fdc4bd115b7b846c7/pkg/kube/util/config/config.go#L30

qu1queee:
> > @edwardstudy an option of what type do you have in mind? you can set it on the `Config` struct, and pass it to the operator `NewManager()` func
> 
> This value is a default, could we set it over cf-operator option.
> 
> https://github.com/cloudfoundry-incubator/cf-operator/blob/c6b81d0a32e7c9fae00e4b0fdc4bd115b7b846c7/pkg/kube/util/config/config.go#L30
 I think this is fair, but we should manage this with another card.

PR #532, opened 2019-08-27
Fixes/improvements

viovanov:


pkg/kube/controllers/boshdeployment/validating_webhook.go
```diff
@@ -81,50 +80,75 @@ func NewValidator(log *zap.SugaredLogger, config *config.Config) admission.Handl
...
-		for {
-			select {
-			case <-timeOut:
-				return false, fmt.Sprintf("Timeout reached. Resource %s does not exist", specOpsResource.Name)
-			case <-tick.C:
-				err := v.client.Get(ctx, key, &corev1.ConfigMap{})
-				if err == nil {
-					return true, fmt.Sprintf("configmap %s, exists", specOpsResource.Name)
-				}
+	for {
```
manno: We're using `wait.Poll` and `wait.PollImmediate` from "k8s.io/apimachinery/pkg/util/wait" in other places. Could be helpful here as well.

PR #533, opened 2019-08-28
Allowed SecurityContext to be passed by Quarks props

f0rmiga:


pkg/bosh/converter/container_factory.go
```diff
@@ -423,7 +424,19 @@ func bpmPreStartInitContainer(
...
+	}
+	if securityContext.Capabilities == nil && len(process.Capabilities) > 0 {
+		securityContext.Capabilities = &corev1.Capabilities{
+			Add: capability(process.Capabilities),
+		}
+	}
+	if securityContext.Privileged == nil {
+		securityContext.Privileged = &process.Unsafe.Privileged
+	}
+	securityContext.RunAsUser = &rootUserID
```
edwardstudy: do we still need to set `securityContext.Privileged` if we specify to run as root user? 
I just saw that `boshPreStartInitContainer()` used `securityContext.RunAsUser = &rootUserID` to replace `Privileged: &privileged,`?

manno: Privileged and root are two different concepts:

* https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privileged (access to host devices)
* https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups

[Line 397](https://github.com/cloudfoundry-incubator/cf-operator/pull/533/files#diff-dac7399b604df4617a12b10ba711fdb8R397) doesn't overwrite, it just adds the rootUID. Since `privileged` was a param before, I expect that to be set in the new param `securityContext`. 
If I remember correctly BOSH pre start containers always need to run as root.

PR #534, opened 2019-08-28
Remove watch on config maps and secrets from ESTS controller

manno:


pkg/kube/controllers/extendedstatefulset/extendedstatefulset_controller.go
```diff
@@ -67,69 +63,5 @@ func AddExtendedStatefulSet(ctx context.Context, config *config.Config, mgr mana
 		return err
 	}
 
-	// Watch ConfigMaps referenced by the ExtendedStatefulSet
-	configMapPredicates := predicate.Funcs{
-		CreateFunc:  func(e event.CreateEvent) bool { return true },
```
manno: Why is this true for CM and false for secrets (line 104)

pkg/kube/util/reference/reconciles.go
```diff
@@ -67,7 +64,7 @@ func GetReconciles(ctx context.Context, client crc.Client, reconcileType Reconci
 			return false, errors.Wrap(err, "error listing references")
 		}
 
-		if versionedSecret && reconcileType != ReconcileForExtendedStatefulSet {
```
manno: Which secrets are left, if we ignore versioned secrets?

pkg/kube/controllers/extendedstatefulset/extendedstatefulset_controller.go
edwardstudy: It should be true both.

PR #535, opened 2019-08-28
Fix updateOnConfigChange cases

edwardstudy:
- Fixed outdated integration case. The controller will bump new versions on config resources change.


edwardstudy:
We should finish this story: https://www.pivotaltracker.com/story/show/168166661
or close this

PR #536, opened 2019-08-28
The BOSH services we create do not need to update on config change

manno:
The only referenced secrets are the desired manifest and instance group
manifest.


manno:
According to @viovanov there could be a reference via Quarks ENV var support, he will come up with a test case :)

PR #539, opened 2019-08-29
Apply CRDs when cf-operator starts

edwardstudy:
[#167425313](https://www.pivotaltracker.com/story/show/167425313)

### DIFF
--- a/cmd/internal/root.go
+++ b/cmd/internal/root.go
@@ -144,7 +144,7 @@ func init() {
 	pf.Int("max-extendedjob-workers", 1, "Maximum of number concurrently running ExtendedJob controller")
 	pf.Int("max-extendedsecret-workers", 5, "Maximum of number concurrently running ExtendedSecret controller")
 	pf.Int("max-extendedstatefulset-workers", 1, "Maximum of number concurrently running ExtendedStatefulSet controller")
-	pf.Bool("apply-crd", true, "If true, apply CRDs on start")
+	pf.Bool("apply-crd", false, "If true, apply CRDs on start")
 	viper.BindPFlag("kubeconfig", pf.Lookup("kubeconfig"))
 	viper.BindPFlag("log-level", pf.Lookup("log-level"))
 	viper.BindPFlag("cf-operator-namespace", pf.Lookup("cf-operator-namespace"))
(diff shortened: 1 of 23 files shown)


pkg/kube/apis/extendedjob/v1alpha1/register.go
```diff
@@ -10,16 +13,116 @@ import (
...
+											Raw: []byte(`"manual"`),
+										},
+										{
+											Raw: []byte(`"once"`),
+										},
+										{
+											Raw: []byte(`"now"`),
+										},
+										{
+											Raw: []byte(`"podstate"`),
```
manno: We no longer support podstate, should have been deleted when trigger ejobs were removed.

pkg/kube/operator/operator.go
```diff
@@ -46,3 +51,75 @@ func NewManager(ctx context.Context, config *config.Config, cfg *rest.Config, op
...
 	return mgr, nil
 }
+
+// ApplyCRDs applies a collection of CRDs into the cluster
+func ApplyCRDs(config *rest.Config) error {
+	exClient, err := extv1client.NewForConfig(config)
+	if err != nil {
+		return errors.Wrap(err, "Could not get kube client")
+	}
+	err = crd.ApplyCRD(
```
manno: This repetition looks like we could put the values in a `[]resource{{name: ..., kind: ...}, {}}` and loop over that, but then it's probably not less code. But if `ApplyCRD` or `WaitForCRDReady` ever change, we only have to fix one invocation.
Well, I have no strong preference here.

edwardstudy: It's better

cmd/internal/root.go
```diff
@@ -129,6 +138,7 @@ func init() {
 	pf.Int("max-extendedjob-workers", 1, "Maximum of number concurrently running ExtendedJob controller")
 	pf.Int("max-extendedsecret-workers", 5, "Maximum of number concurrently running ExtendedSecret controller")
 	pf.Int("max-extendedstatefulset-workers", 1, "Maximum of number concurrently running ExtendedStatefulSet controller")
+	pf.Bool("apply-crd", true, "If true, apply CRDs on start")
```
manno: false so we are backwards compatible for now

pkg/kube/apis/extendedjob/v1alpha1/register.go
edwardstudy: done

pkg/kube/operator/operator.go
edwardstudy: done

cmd/internal/root.go
edwardstudy: done

PR #546, opened 2019-08-30
Use an external IP to make the webhook accessible on GKE

manno:
For the hands on lab session on GKE


manno:
Closed in favor of #552

PR #547, opened 2019-09-02
Fixes and improvements for getting Eirini to work on scf v3

viovanov:
- fixes errands and auto-errands (they weren't being run before)
- improves error management on pre-render scripts (stdout and stderr was lost)
- add support for referencing a service in an ExtendedSecret certificate definition; the IPs and hostnames of that service are added as SANs


pkg/bosh/converter/bpm_resources.go
```diff
@@ -349,8 +354,9 @@ func (kc *KubeConverter) errandToExtendedJob(
 			Annotations: instanceGroup.Env.AgentEnvBoshConfig.Agent.Settings.Annotations,
 		},
 		Spec: ejv1.ExtendedJobSpec{
+
```
f0rmiga: [nit] This extra line seems odd here.

pkg/bosh/converter/bpm_resources.go
```diff
@@ -340,6 +340,11 @@ func (kc *KubeConverter) errandToExtendedJob(
 	volumes = append(volumes, defaultVolumes...)
 	volumes = append(volumes, bpmVolumes...)
 
+	strategy := ejv1.TriggerManual
+	if instanceGroup.LifeCycle == "auto-errand" {
```
f0rmiga: Ideally, we should transform these hardcoded strings into constants.

pkg/bosh/manifest/cmd_render_job_tmpls.go
```diff
@@ -152,54 +152,55 @@ func runRenderScript(
...
-			go func() {
-				for outScanner.Scan() {
-					fmt.Printf("pre-render-out-%d | %s\n", idx, outScanner.Text())
-				}
-			}()
-		}
+		errScanner := bufio.NewScanner(errReader)
+		go func() {
+			for errScanner.Scan() {
+				errBuffer.Write([]byte(fmt.Sprintf("%s\n", errScanner.Text())))
```
f0rmiga: We should take advantage of `fmt.Fprintf` here and below:
```go
fmt.Fprintf(&errBuffer, "%s\n", errScanner.Text())
```

pkg/bosh/manifest/cmd_render_job_tmpls.go
```diff
@@ -152,54 +152,55 @@ func runRenderScript(
...
+			return createErr(err)
+		}
 
-			errScanner := bufio.NewScanner(errReader)
-			go func() {
-				for errScanner.Scan() {
-					fmt.Printf("pre-render-err-%d | %s\n", idx, errScanner.Text())
-				}
-			}()
+		outBuffer := bytes.NewBufferString("")
```
f0rmiga: Using `strings.Builder` is preferable here. You can simply:
```go
var outBuffer, errBuffer strings.Builder
```
It will initialize the buffers correctly.

pkg/bosh/manifest/cmd_render_job_tmpls.go
```diff
@@ -129,7 +130,6 @@ func runRenderScript(
 	scriptType string,
 	scripts []string,
 	igName string,
-	silent bool,
```
f0rmiga: +1

pkg/kube/controllers/extendedjob/errand_reconciler.go
```diff
@@ -99,3 +116,41 @@ func (r *ErrandReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	return reconcile.Result{}, nil
 }
+
+// injectContainerEnv injects AZ info to container envs
+// errands always have an AZ_INDEX of 1
+func (r *ErrandReconciler) injectContainerEnv(podSpec *corev1.PodSpec) {
+
+	containers := []*corev1.Container{}
```
f0rmiga: Please, initialize the slice with capacity to avoid unnecessary memory copying:
```go
containers := make([]*corev1.Container, 0, len(podSpec.Containers) + len(podSpec.InitContainers))
```

pkg/kube/controllers/extendedjob/errand_reconciler.go
```diff
@@ -99,3 +116,41 @@ func (r *ErrandReconciler) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		containers = append(containers, &podSpec.Containers[i])
+	}
+	for i := 0; i < len(podSpec.InitContainers); i++ {
+		containers = append(containers, &podSpec.InitContainers[i])
+	}
+	for _, container := range containers {
+		envs := container.Env
+
+		// Default to zone 1, with 1 replica
+		envs = upsertEnvs(envs, EnvCfOperatorAzIndex, "1")
```
f0rmiga: We can create `envs` outside the `for` block and the `for` block can become:
```go
for _, container := range containers {
  container.Env = envs.DeepCopy()
}
```

This is better due to the upsert being called repeatedly.

viovanov: Not all container envs are the same

f0rmiga: Oh, I completely missed the fact that we are grabbing the current envs. :expressionless:

PR #548, opened 2019-09-02
Remove watch on secret update from BPM controller

manno:
BPM controller looks for IG manifests and BPM configs, both are
versioned secrets. Versioned secrets are not mutated, a new one with an
increased version (as part of the name) is created for every change.

Therefore it's not necessary to watch a versioned secret for updates. Unless they are 'user controlled' maybe?


edwardstudy:
It makes sense

PR #551, opened 2019-09-02
Improve logging and comments

manno:


pkg/kube/controllers/boshdeployment/deployment_controller.go
```diff
@@ -42,7 +43,25 @@ func AddDeployment(ctx context.Context, config *config.Config, mgr manager.Manag
 	}
 
 	// Watch for changes to primary resource BOSHDeployment
-	err = c.Watch(&source.Kind{Type: &bdv1.BOSHDeployment{}}, &handler.EnqueueRequestForObject{})
+	p := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			ctxlog.NewPredicateEvent(e.Object).Debug(
+				ctx, e.Meta, "bdv1.BOSHDeployment",
+				fmt.Sprintf("BOSHDeployment's create predicate passed for %s", e.Meta.GetName()),
```
edwardstudy: How about this? `BOSHDeployment '%' creation predicate passed`

And these logs can be indicated source kind:
https://github.com/cloudfoundry-incubator/cf-operator/blob/f3967fda1c09d694f9ac0daa5c996022e5750870/pkg/kube/controllers/boshdeployment/bpm_controller.go#L53
https://github.com/cloudfoundry-incubator/cf-operator/blob/f3967fda1c09d694f9ac0daa5c996022e5750870/pkg/kube/controllers/boshdeployment/generated_variable_controller.go#L54

manno: I like that your suggestion is more terse, but for me 'create' always was a thing: `CreateFunc: {}`.  To me, 'creation predicate' sounds too much like the predicate is creating something.

The same wording is used in other controllers and for update, too.

PR #552, opened 2019-09-02
Add provider flag for GKE

rohitsakala:
[#168103576](https://www.pivotaltracker.com/n/projects/2192232/stories/168103576)

### DIFF
--- a/cmd/internal/root.go
+++ b/cmd/internal/root.go
@@ -28,6 +28,8 @@ import (
 
 const (
 	cfFailedMessage = "cf-operator command failed."
+	// Port on which the controller-runtime manager listens
+	managerPort = 2999
 )
 
 var (
@@ -60,17 +62,20 @@ var rootCmd = &cobra.Command{
 		log.Infof("Starting cf-operator %s with namespace %s", version.Version, cfOperatorNamespace)
 		log.Infof("cf-operator docker image: %s", converter.GetOperatorDockerImage())
 
-		host := viper.GetString("operator-webhook-service-host")
-		port := viper.GetInt32("operator-webhook-service-port")
+		serviceHost := viper.GetString("operator-webhook-service-host")
+		// Port on which the cf operator webhook kube service listens to.
+		servicePort := viper.GetInt32("operator-webhook-service-port")
+		provider := viper.GetString("provider")
 
-		if host == "" {
+		if serviceHost == "" && provider == "" {
 			log.Fatal("%s operator-webhook-service-host flag is not set (env variable: CF_OPERATOR_WEBHOOK_SERVICE_HOST).", cfFailedMessage)
 		}
 
 		config := config.NewConfig(
 			cfOperatorNamespace,
-			host,
-			port,
+			provider,
+			serviceHost,
+			servicePort,
 			afero.NewOsFs(),
 			viper.GetInt("max-boshdeployment-workers"),
 			viper.GetInt("max-extendedjob-workers"),
@@ -83,7 +88,7 @@ var rootCmd = &cobra.Command{
 			Namespace:          cfOperatorNamespace,
 			MetricsBindAddress: "0",
 			LeaderElection:     false,
-			Port:               int(port),
+			Port:               managerPort,
 			Host:               "0.0.0.0",
 		})
 		if err != nil {
@@ -125,6 +130,7 @@ func init() {
 	pf.StringP("operator-webhook-service-host", "w", "", "Hostname/IP under which the webhook server can be reached from the cluster")
 	pf.StringP("operator-webhook-service-port", "p", "2999", "Port the webhook server listens on")
 	pf.StringP("docker-image-tag", "t", version.Version, "Tag of the operator docker image")
+	pf.StringP("provider", "x", "", "Cloud Provider where cf-operator is being deployed")
 	pf.Int("max-boshdeployment-workers", 1, "Maximum of number concurrently running BOSHDeployment controller")
 	pf.Int("max-extendedjob-workers", 1, "Maximum of number concurrently running ExtendedJob controller")
 	pf.Int("max-extendedsecret-workers", 5, "Maximum of number concurrently running ExtendedSecret controller")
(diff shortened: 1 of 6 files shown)


cmd/internal/root.go
```diff
@@ -62,13 +62,15 @@ var rootCmd = &cobra.Command{
 
 		host := viper.GetString("operator-webhook-service-host")
 		port := viper.GetInt32("operator-webhook-service-port")
+		provider := viper.GetString("provider")
 
-		if host == "" {
+		if host == "" && provider == "" {
```
edwardstudy: nit: better error message for empty provider

cmd/internal/root.go
```diff
@@ -83,7 +85,7 @@ var rootCmd = &cobra.Command{
 			Namespace:          cfOperatorNamespace,
 			MetricsBindAddress: "0",
 			LeaderElection:     false,
-			Port:               int(port),
+			Port:               2999,
```
edwardstudy: default port is 2999, right? why need to hardcoded here?

deploy/helm/cf-operator/values.yaml
```diff
@@ -23,3 +23,5 @@ serviceAccount:
   cfOperatorServiceAccount:
     create: true
     name:
+
+provider:
```
edwardstudy: nit: empty string ""

pkg/kube/controllers/webhook_configuration.go
```diff
@@ -166,27 +172,47 @@ func (f *WebhookConfig) generateValidationWebhookServerConfig(ctx context.Contex
...
 	}
 
-	for _, webhook := range webhooks {
-		ctxlog.Debugf(ctx, "Calculating validation webhook '%s'", webhook.Name)
-
-		url := url.URL{
-			Scheme: "https",
-			Host:   net.JoinHostPort(f.config.WebhookServerHost, strconv.Itoa(int(f.config.WebhookServerPort))),
-			Path:   webhook.Path,
+	if f.config.Provider == "gke" {
```
edwardstudy: could we only create `admissionregistrationv1beta1.WebhookClientConfig` for different cases and then assign it to webhook.ClientConfig?

cmd/internal/root.go
rohitsakala: It is actually not for error message. `provider` is optional. host is necessary if provider is not given. If provider is given, then host becomes optional as we use service address for webhook.

cmd/internal/root.go
rohitsakala: When installing using Helm, the port specified here [0] is taken as the value. It happens like this because helm installation creates a service which puts env variable `CF_OPERATOR_WEBHOOK_SERVICE_PORT` (2999) [0] and the viper takes this as port. This is wrong as we should give target port of the service to start the manager. so I hardcoded it for now. 

[0] https://github.com/cloudfoundry-incubator/cf-operator/blob/master/deploy/helm/cf-operator/templates/service.yaml#L9

deploy/helm/cf-operator/values.yaml
rohitsakala: Done.

pkg/kube/controllers/webhook_configuration.go
rohitsakala: I wanted to do that, but the `Path` key in `WebhookClientConfig` depends on the for loop. So couldn't do that.

cmd/internal/root.go
manno: Can we put this in a constant so the name explains why it's hardcoded? `servicePort`, maybe? Hm, or maybe put a comment in the code to explain the difference between the two ports.

So one is for connecting to the k8s service, the other for the controller-runtime manager to open a listen socket for the webhook server?

pkg/kube/controllers/webhook_configuration.go
manno: Why, for performance reasons? I don't think creating a struct is especially expensive  in Go. I should benchmark that sometime...

Oh, because of the duplication with the else branch?
I'd try to  introduce a `func newWebhook(webhook, clientConfig) admissionregistrationv1beta1.Webhook` to dry this out

cmd/internal/root.go
rohitsakala: Done.

pkg/kube/controllers/webhook_configuration.go
rohitsakala: Done. Thanks :)

pkg/kube/controllers/webhook_configuration.go
```diff
@@ -169,26 +175,31 @@ func (f *WebhookConfig) generateValidationWebhookServerConfig(ctx context.Contex
...
-		}
-		urlString := url.String()
-		wh := admissionregistrationv1beta1.Webhook{
-			Name:              webhook.Name,
-			Rules:             webhook.Rules,
-			FailurePolicy:     &webhook.FailurePolicy,
-			NamespaceSelector: webhook.NamespaceSelector,
-			ClientConfig: admissionregistrationv1beta1.WebhookClientConfig{
+		if f.config.Provider == "gke" {
+			clientConfig := admissionregistrationv1beta1.WebhookClientConfig{
```
satadruroy: I suspect you might need to do the same for EKS (we saw the same problem on EKS). Deleting the webhook made it work so under the hood seems like the EKS apiserver also cannot contact the webhook service using the clusterip/port.

jandubois: Having provider-specific logic hard-coded in the Go code seems pretty inflexible to me and creates a hard dependency between operator release and provider version.

I think it would be better if the different logic would be configured via some property, and that property could be overridden at deployment time. So it could be determined (if possible) by helm, or otherwise is specified by the deployer.

Anyways, the feature modification should be triggered by some underlying feature name and not by a provider name.

satadruroy: Is there any reason why we should not use the `service` field in all cases and not special case it? The kube documentation seems to suggest so unless I'm mis-reading it?

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#url

**The host should not refer to a service running in the cluster**; use a service reference by specifying the service field instead.

PR #556, opened 2019-09-04
Provide factory interfaces

edwardstudy:
One of [#167507451](https://www.pivotaltracker.com/story/show/167507451)

- Refactor EJob factory signatures
- Remove job label of manifest SHA1
- Add jobFactory unit cases for deployment reconciler
- Add container factory interface
- Add kubeConverter interface


pkg/bosh/converter/container_factory.go
```diff
@@ -24,7 +24,13 @@ const (
 )
 
 // ContainerFactory builds Kubernetes containers from BOSH jobs.
-type ContainerFactory struct {
+type ContainerFactory interface {
```
manno: Can we follow the advice from https://github.com/golang/go/wiki/CodeReviewComments#interfaces and declare the interface on the consumer side? The decoupling is nice and we see directly which methods from the interface are used by every consumer.

If you decide to keep the interface here, we need a check like `var _ reconcile.Reconciler = &ReconcileBOSHDeployment{}`. These seem to be missing for several of our interfaces :?

pkg/bosh/converter/container_factory.go
```diff
@@ -24,7 +24,13 @@ const (
...
+	JobsToInitContainers(jobs []bdm.Job, defaultVolumeMounts []corev1.VolumeMount, bpmDisks BPMResourceDisks) ([]corev1.Container, error)
+	JobsToContainers(jobs []bdm.Job, defaultVolumeMounts []corev1.VolumeMount, bpmDisks BPMResourceDisks) ([]corev1.Container, error)
+}
+
+// ConcreteContainerFactory is a concrete implementation of ContainerFactor.
+type ConcreteContainerFactory struct {
```
manno: `ContainerFactoryImpl` seems to be a common naming in Golang (Java?). While the `Impl` suffix doesn't add much, I don't think 'Concrete', 'Base' or 'Default' are any better ;)

This naming problem goes away if the the interface is declared in the consumer's package. Otherwise, can we find a more descriptive name? "K8SContainerFactory", "NotFakeContainerFactory", ugh no, apparently I can't :)
We only add it for mocking anyhow and I don't see how there could be another implementation of that factory.

pkg/bosh/converter/container_factory.go
```diff
@@ -33,9 +39,9 @@ type ContainerFactory struct {
 	bpmConfigs           bpm.Configs
 }
 
-// NewContainerFactory returns a new ContainerFactory for a BOSH instant group.
-func NewContainerFactory(manifestName string, instanceGroupName string, version string, disableLogSidecar bool, releaseImageProvider ReleaseImageProvider, bpmConfigs bpm.Configs) *ContainerFactory {
-	return &ContainerFactory{
+// NewConcreteContainerFactory returns a concrete implementation of ContainerFactory.
+func NewConcreteContainerFactory(manifestName string, instanceGroupName string, version string, disableLogSidecar bool, releaseImageProvider ReleaseImageProvider, bpmConfigs bpm.Configs) ContainerFactory {
```
manno: There are advantages of returning a concrete struct here, instead of the interface: https://mycodesmells.com/post/accept-interfaces-return-struct-in-go

pkg/bosh/converter/job_factory.go
```diff
@@ -38,75 +38,69 @@ const (
 )
 
 // JobFactory creates Jobs for a given manifest
-type JobFactory struct {
-	Manifest            bdm.Manifest
-	Namespace           string
-	desiredManifestName string
+type JobFactory interface {
+	VariableInterpolationJob(manifest bdm.Manifest) (*ejv1.ExtendedJob, error)
```
manno: :+1: moving this into the parameters makes the factory more resilient.

pkg/bosh/converter/job_factory.go
```diff
@@ -150,40 +144,44 @@ func (f *JobFactory) VariableInterpolationJob() (*ejv1.ExtendedJob, error) {
 }
 
 // InstanceGroupManifestJob generates the job to create an instance group manifest
-func (f *JobFactory) InstanceGroupManifestJob() (*ejv1.ExtendedJob, error) {
-	containers := make([]corev1.Container, len(f.Manifest.InstanceGroups))
+func (f *ConcreteJobFactory) InstanceGroupManifestJob(manifest bdm.Manifest) (*ejv1.ExtendedJob, error) {
+	containers := make([]corev1.Container, len(manifest.InstanceGroups))
+
+	// ExtendedJob will always pick the latest version for versioned secrets
+	desiredManifestName := names.DesiredManifestName(manifest.Name, "1")
```
manno: In other parts of the code base I see a `names.DesiredManifestName(manifest.Name, "")`, so sometimes we want a specific version, sometimes we just give some version ("1") and when we give "" we just want the manifest name.

pkg/bosh/converter/container_factory.go
edwardstudy: Thx. I decided to move interfaces to consumer's package.

pkg/bosh/converter/job_factory.go
edwardstudy: If you use `names.DesiredManifestName(manifest.Name, "")`, it's mean that you get secret prefix and then use `versionedSecretStore.Latest` to fetch the latest desired manifest.

In this case, it's **another** path to set resources' references that contains versioned desired manifest like placeholder, and let `GetReconciles` func knows this object has versioned secret and determine current versioned secret is referenced by this object. 

Details info is in `reference` package: `pkg/kube/util/reference/reconciles.go`

edwardstudy:
@manno Hi, Mario. Thx for the review. My feedback is ready now.

pkg/kube/controllers/extendedjob/job_reconciler.go
```diff
@@ -1,6 +1,7 @@
 package extendedjob
 
 import (
+	"code.cloudfoundry.org/cf-operator/pkg/bosh/factory"
```
manno: move into import group below

edwardstudy: oops

edwardstudy:
> Some imports are on top of the list, should be grouped with the rest.
> 
> I'd not use a factory package, but keep everything in `converter`, because the factories are only used there and we want to avoid too generic and helper packages like `util` and such. (Which we already have.. haha)
> However if there is a good reason for that 'factory' package, then we could drop the `_factory` suffix from the files, as they are already in that folder. And from the structs, too, because `factory.ContainerFactory` is redundant?

I accept

> Some imports are on top of the list, should be grouped with the rest.
> 
> I'd not use a factory package, but keep everything in `converter`, because the factories are only used there and we want to avoid too generic and helper packages like `util` and such. (Which we already have.. haha)
> However if there is a good reason for that 'factory' package, then we could drop the `_factory` suffix from the files, as they are already in that folder. And from the structs, too, because `factory.ContainerFactory` is redundant?

I extracted factories because their consumer `kubeConverter` is also in `converter`. So it's

edwardstudy:
> Some imports are on top of the list, should be grouped with the rest.
> 
> I'd not use a factory package, but keep everything in `converter`, because the factories are only used there and we want to avoid too generic and helper packages like `util` and such. (Which we already have.. haha)
> However if there is a good reason for that 'factory' package, then we could drop the `_factory` suffix from the files, as they are already in that folder. And from the structs, too, because `factory.ContainerFactory` is redundant?

Split factory from converter package is because their consumer `kubeConverter` is in there.

aduffeck:
+1 for keeping the factories in the `converter` package. They are not meant to be used outside of it but are merely a means of making the converter more testable.
Keeping them in a separate package could raise wrong expectations or lead to a mix of different scopes if it is extended with other factories in the future.

PR #559, opened 2019-09-12
Copy packages to shared volume and mount as /var/vcap/packages

f0rmiga:
For each release, an init container will copy the packages into a shared ephemeral volume that later gets mounted as `/var/vcap/packages` in the main containers.

It solves the problem of jobs using packages from other jobs in the same instance group and reduces the number of patches required on SCF to make buildpacks and smoke-tests work.

### DIFF
--- a/pkg/bosh/converter/container_factory.go
+++ b/pkg/bosh/converter/container_factory.go
@@ -287,7 +287,7 @@ func containerRunCopier() corev1.Container {
 func jobSpecCopierContainer(releaseName string, jobImage string, volumeMountName string) corev1.Container {
 	inContainerReleasePath := filepath.Join(VolumeRenderingDataMountPath, "jobs-src", releaseName)
 	return corev1.Container{
-		Name:  names.Sanitize(fmt.Sprintf("spec-copier-%s", releaseName)),
+		Name:  names.Sanitize("spec-copier-%s", releaseName),
 		Image: jobImage,
 		VolumeMounts: []corev1.VolumeMount{
 			{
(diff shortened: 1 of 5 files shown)


viovanov:
This seems like a bad idea because of the following:

- upgrades (how do we keep the volume in sync)
- this adds to storage needs
- solves a problem that's been solved (docker images)
- seems like a workaround for bad actors (we should just fix the bad actors)

f0rmiga:
@viovanov 

- I don't see why it affects upgrades as the volume is ephemeral and re-populated on every pod creation.
- True, and this is the major issue that held me from opening a normal PR instead of a draft one.
- Not quite true, we cannot mount docker images as volumes inside another container.
- True. Is it possible to fix the bad actors before we consume them?

jandubois:
@viovanov I had some of the same objections when I got notified about this PR and talked to @f0rmiga about it in person.

I hadn't realized that sometimes bosh uses release co-location to provide additional binaries to a job in a different release. Some examples are:

* buildpacks come in their own releases, are installed by a CC job
* rootfs come releases, used by diego runner
* bpm binary has its own release, used by almost all other release jobs now
* cfdot binary has its own release, used by multiple diego instance-groups

The last time I looked there was no mechanism in k8s to mount the fs from a different container; the closest I ever found were flexvolume drivers, but that would require a custom driver that must be installed at the kube level, and the only implementation I found was a quick proof-of-concept.

So right now I agree with Thulio that this is currently needed as a workaround until kube provides a more native mechanism.

viovanov:
@jandubois yes, there are cases where bosh releases ignore the BOSH contract
e.g.: the proper way would be to copy buildpacks to `/var/vcap/data`, where the CC can look for them

We should talk to @ericpromislow and fix this upstream.

jandubois:
> @jandubois yes, there are cases where bosh releases ignore the BOSH contract
> e.g.: the proper way would be to copy buildpacks to `/var/vcap/data`, where the CC can look for them

I agree that it would be better if colocated jobs would copy the data in a `pre-start` script to the shared data volume.

But the bosh documentation doesn't mandate this:
https://bosh.io/docs/package-vendoring/#colocation

It actually gives an example that shows that a merged filesystem is expected:

>BPM release making `/var/vcap/jobs/bpm/bin/bpm` available to all other releases...

So I'm all for changing this upstream, but I'm not sure if they will accept that the current practice "ignores the bosh contract".

f0rmiga:
@viovanov @jandubois Closing this for now as we will find better alternatives to fix the issue.

PR #563, opened 2019-09-16
Helm notes fix namespace name

rohitsakala:


deploy/helm/cf-operator/templates/NOTES.txt
```diff
@@ -37,6 +37,6 @@ Interacting with the cf-operator pod
   kubectl -n {{ .Release.Namespace }} create -f docs/examples/bosh-deployment/boshdeployment-with-custom-variable.yaml
 
 4. See the cf-operator in action!
-  watch -c "kubectl -n cf-operator2 get pods"
+  watch -c "kubectl -n cf-operator get pods"
```
aduffeck: I think this has to be

```suggestion
  watch -c "kubectl -n {{ .Release.Namespace }} get pods"
```

to make it work generically?

PR #566, opened 2019-09-16
ExtendedSecret should not reconcile for status changes

manno:
[#166767434](https://www.pivotaltracker.com/story/show/166767434)


manno:
> Do we have any test case for this?

No. 
We still have the story about unit testing controllers and reconcilers in the ice box.
I'm also not sure how to integration test this, we want to avoid negative tests ("something doesn't happen"), because they take so long.

However I'm currently testing this manually by deploying SCFv3.

PR #574, opened 2019-09-19
allow updation of status field in exsec

rohitsakala:


pkg/kube/controllers/extendedsecret/extendedsecret_reconciler_test.go
```diff
@@ -333,18 +333,24 @@ var _ = Describe("ReconcileExtendedSecret", func() {
...
 			}
 
-			client.CreateCalls(func(context context.Context, object runtime.Object, _ ...crc.CreateOption) error {
-				secret := object.(*corev1.Secret)
-				Expect(secret.StringData["password"]).To(Equal(password))
-				Expect(secret.GetName()).To(Equal("generated-secret"))
-				Expect(secret.GetLabels()).To(HaveKeyWithValue(esv1.LabelKind, esv1.GeneratedSecretKind))
+			client.UpdateCalls(func(context context.Context, object runtime.Object, _ ...crc.UpdateOption) error {
+				switch object.(type) {
+				case *esv1.ExtendedSecret:
```
manno: Like staticcheck suggest:

pkg/kube/controllers/extendedsecret/extendedsecret_reconciler_test.go:337:12: assigning the result of this type assertion to a variable (switch object := object.(type)) could eliminate the following type assertions:

pkg/kube/util/mutate/mutate.go
```diff
@@ -71,7 +71,7 @@ func ESecMutateFn(eSec *esv1.ExtendedSecret) controllerutil.MutateFn {
 		eSec.Labels = updated.Labels
 		eSec.Annotations = updated.Annotations
 		// Update only when spec has been changed
-		if !reflect.DeepEqual(eSec.Spec, updated.Spec) {
+		if !reflect.DeepEqual(eSec.Spec, updated.Spec) || eSec.Status.Generated != updated.Status.Generated {
```
manno: this could be two separate `if`s. There is no reason to copy spec, if just status changed and vice versa.

pkg/kube/controllers/extendedsecret/extendedsecret_reconciler_test.go
rohitsakala: Done.

pkg/kube/util/mutate/mutate.go
rohitsakala: Done.

PR #577, opened 2019-09-19
Patch webhooks instead of deleting

jkbschmid:
To avoid having to delete the webhooks after a crash, we tried to integrate an `objectSelector` which excludes the `cf-operator` pod from the `MutatingWebhook`. However, this is a larger change because it requires updating to a newer version of `k8s.io/apimachinery`. As a first step, we added it to the README as a patch.

Do you think adding an exception for cf-operator to the webhook is a good idea? If so, we could try to provide a PR that does this directly in the go-code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,12 +35,19 @@ For more information about the `cf-operator` helm chart and how to configure it,
 
 ### Recovering from a crash
 
-If the operator pod crashes, it cannot be restarted in the same namespace before the existing mutating webhook configuration for that namespace is temporarily patched.
+If the operator pod crashes, it cannot be restarted in the same namespace before the existing mutating webhook configuration for that namespace is removed.
 The operator uses mutating webhooks to modify pods on the fly and Kubernetes fails to create pods if the webhook server is unreachable.
 The webhook configurations are installed cluster wide and don't belong to a single namespace, just like custom resources.
 
-To patch the webhook configurations for the cf-operator namespace run:
+To remove the webhook configurations for the cf-operator namespace run:
 
+```bash
+CF_OPERATOR_NAMESPACE=cf-operator
+kubectl delete mutatingwebhookconfiguration "cf-operator-hook-$CF_OPERATOR_NAMESPACE"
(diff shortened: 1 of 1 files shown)


phil9909:
> What's the benefit of this change? We still need to change `mutatingwebhookconfiguration` after the crash

For the UX it does not really make a difference, yes.

But it certainly is a small improvement: Deleting the webhook means that now all pods can be restarted and the webhook will only fire once `cf-operator` is fully booted and has recreated the webhook. On the otherhand, using this solution only the `cf-operator` pod can start and only after the operator is started and responds to webhooks again the other pods will start.  (I don't know if this even matters.) TL;DR we don't miss any events.

Our idea behind this change was: It is one step closer to a real solution and at least it's now documented how it could work, once dependencies are updated.


BTW: I signed the CLA. How can I trigger the @cfdreddbot to check again?

aduffeck:
Please note that there's already a story in tracker to exclude the operator pod from the webhooks (https://www.pivotaltracker.com/story/show/167425277). We decided to postpone it for a while, though because it requires kubernetes 1.15 which is pretty new and we don't want to depend on it yet.

For the same reason I think we should rather not refer to 1.15+ features in the documentation.

@phil9909 In order to re-trigger cfdreddbot you have to close and re-open the pr afaik.

manno:
Maybe the time to require Kubernetes 1.15 is now, it's already 3 month old ;)

Anyhow, before we can solve this in code, it might be helpful improve our 'recovering from crash' documentation.

@aduffeck  would it be agreeable if the new text was in an explicit "if you are on Kubernetes 1.15" section?

aduffeck:
:+1: having it as an additional section specific to 1.15+ environments sounds good to me.

jkbschmid:
Good to know there is already a story for solving the problem with the webhooks:)
@aduffeck Could you please give us access to the 'test' logs, so we can check whether the test fails because of our PR? I get a 401 unauthorized when trying to access them.

aduffeck:
@jkbschmid we're currently trying to figure out how to solve this logs issue best. We'll let you know...

**Spoiler**: Tests failed because of issues with the cloud environment, not because of your changes :)

PR #584, opened 2019-09-24
Test for zero-downtime upgrades

aduffeck:
[#165014929](https://www.pivotaltracker.com/story/show/165014929)


integration/deploy_test.go
```diff
@@ -251,16 +253,67 @@ var _ = Describe("Deploy", func() {
...
 					cm, err := env.GetConfigMap(env.Namespace, "manifest")
 					Expect(err).NotTo(HaveOccurred())
 					cm.Data["manifest"] = strings.Replace(cm.Data["manifest"], "changeme", "dont", -1)
 					_, _, err = env.UpdateConfigMap(env.Namespace, *cm)
 					Expect(err).NotTo(HaveOccurred())
 
-					By("checking for instance group updated pods")
 					err = env.WaitForInstanceGroup(env.Namespace, "test", "nats", "2", 2)
 					Expect(err).NotTo(HaveOccurred(), "error waiting for instance group pods from deployment")
+
```
viovanov: I think we should also wait for `nats-v1` to go away.

aduffeck: Good point, I extended the test accordingly.

PR #585, opened 2019-09-24
Add error for invalid cf-operator docker image reference

manno:
When the command line args for cf-operators docker image source are invalid, throw an error.

Also some clean ups, which I needed to get `make test` run locally. 

[#167509239](https://www.pivotaltracker.com/story/show/167509239)

### DIFF
--- a/pkg/kube/util/names/docker_source.go
+++ b/pkg/kube/util/names/docker_source.go
@@ -1,0 +1,27 @@
+package names
+
+import (
+	"errors"
+	"path"
+)
+
+// GetDockerSourceName returns the name of the docker image
+// More info:
+//   * https://kubernetes.io/docs/concepts/containers/images
+//   * <hub-user>/<repo-name>[:<tag>]
+//   * ACCOUNT.dkr.ecr.REGION.amazonaws.com/imagename:tag
+//
+// prefix: [<host>[:<port>]/][<org>/]
+// repo: <name>
+// tag :[:<tag>]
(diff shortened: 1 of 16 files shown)


pkg/kube/util/names/docker_source.go
```diff
@@ -0,0 +1,27 @@
...
+
+import "errors"
+
+// GetDockerSourceName returns the name of the docker image
+// More info:
+//   * https://kubernetes.io/docs/concepts/containers/images
+//   * <hub-user>/<repo-name>[:<tag>]
+//   * ACCOUNT.dkr.ecr.REGION.amazonaws.com/imagename:tag
+//
+// prefix: [<host>[:<port>]/][<repo>/]
```
aduffeck: Should this be 
```suggestion
// prefix: [<host>[:<port>]/][<org>/]
```
?

pkg/kube/util/names/docker_source.go
```diff
@@ -0,0 +1,27 @@
...
+// repo: <name>
+// tag :[:<tag>]
+func GetDockerSourceName(prefix, repo, tag string) (string, error) {
+	source := ""
+	if repo == "" {
+		return "", errors.New("repository is mandatory")
+	}
+	if prefix != "" {
+		source = prefix + "/"
```
aduffeck: The docs above say that `prefix` should already end with a `/`. Maybe use `path.Join()` instead to make it robust?

manno: :+1:

PR #598, opened 2019-09-30
Bosh dns core dns

kramerul:
We implemented [BOSH-compatible hostname aliases](https://github.com/SUSE/scf/issues/2780)

* We use a second DNS server
* k8s headless services are created based on the alias configuration
* The _old_ headless service names are resolved by the second DNS server.

All these features are only enabled if the bosh dns addon is configured inside the manifest.

It can be tested using `docs/examples/bosh-deployment/boshdeployment-with-bosh-dns.yaml`.

It's also working with the [SUSE cloudfoundry v3-release](https://github.com/SUSE/scf) after changing the file `scripts/scf/bazel-scf/deploy/helm/scf/assets/operations/addons.yaml`
* Remove 
```yaml
- type: remove
  path: /addons/name=bosh-dns-aliases
```

* Add*

```yaml
- type: replace
  path: /addons/name=bosh-dns-aliases/jobs/name=bosh-dns-aliases/properties/aliases/domain=nats?
  value:
    domain: nats.service.cf.internal
    targets:
      - deployment: cf
        domain: bosh
        instance_group: nats
        network: default
        query: '*'
- type: replace
  path: /addons/name=bosh-dns-aliases/jobs/name=bosh-dns-aliases/properties/aliases/domain=eirini?
  value:
    domain: eirini.service.cf.internal
    targets:
      - deployment: cf
        domain: bosh
        instance_group: eirini
        network: default
        query: '*'
```

### DIFF
--- a/cmd/internal/bpm_configs.go
+++ b/cmd/internal/bpm_configs.go
@@ -3,9 +3,11 @@ package cmd
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
+	"runtime/debug"
 	"time"
 
 	"github.com/pkg/errors"
@@ -30,6 +32,16 @@ instance group.
 	RunE: func(cmd *cobra.Command, args []string) (err error) {
 		defer func() {
 			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+				fmt.Fprintf(os.Stderr, "Sleeping ...\n")
+				time.Sleep(debugGracePeriod)
+			}
+		}()
+
+		defer func() {
+			if r := recover(); r != nil {
+				fmt.Fprintf(os.Stderr, "Recovered in f: %v\n%s\n", r, string(debug.Stack()))
+				fmt.Fprintf(os.Stderr, "Sleeping ...\n")
 				time.Sleep(debugGracePeriod)
 			}
 		}()
--- a/cmd/internal/bpm_configs.go
+++ b/cmd/internal/bpm_configs.go
@@ -31,16 +31,11 @@ instance group.
 `,
 	RunE: func(cmd *cobra.Command, args []string) (err error) {
 		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("recovered in f: %v\n%s", r, string(debug.Stack()))
+			}
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-				fmt.Fprintf(os.Stderr, "Sleeping ...\n")
-				time.Sleep(debugGracePeriod)
-			}
-		}()
-
-		defer func() {
-			if r := recover(); r != nil {
-				fmt.Fprintf(os.Stderr, "Recovered in f: %v\n%s\n", r, string(debug.Stack()))
 				fmt.Fprintf(os.Stderr, "Sleeping ...\n")
 				time.Sleep(debugGracePeriod)
 			}
--- a/pkg/bosh/manifest/dns.go
+++ b/pkg/bosh/manifest/dns.go
@@ -5,14 +5,11 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
-	"io/ioutil"
 	"regexp"
 	"strings"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
-
-	"code.cloudfoundry.org/cf-operator/pkg/kube/util/ctxlog"
 
 	appsv1 "k8s.io/api/apps/v1"
 
@@ -222,27 +219,6 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
 	return nil
 }
 
-var re = regexp.MustCompile(`nameserver\s+([\d\.]*)`)
-
-func (dns *boshDomainNameService) rootDNSIP() string {
-	content, err := ioutil.ReadFile("/etc/resolv.conf")
-	if err != nil {
-		content = []byte{}
-	}
-	return GetNameserverFromResolveConfig(content)
-}
-
-// GetNameserverFromResolveConfig read nameserver from resolve.conf
-func GetNameserverFromResolveConfig(content []byte) string {
-	found := re.FindSubmatch(content)
-	if len(found) == 0 {
-		ctxlog.Errorf(context.Background(), "No nameserver found in resolve.conf using 1.1.1.1")
-		return "1.1.1.1"
-	}
-	return string(found[1])
-
-}
-
 type simpleDomainNameService struct {
 	ManifestName string
 }
@@ -327,7 +303,7 @@ func createCorefile(dns *boshDomainNameService, namespace string) string {
     health
 %s
     rewrite name substring service.cf.internal. %s.svc.cluster.local.
-    forward . %s
+    forward . /etc/resolv.conf
 }
-	`, rewrites, namespace, dns.rootDNSIP())
+	`, rewrites, namespace)
 }
--- a/pkg/bosh/manifest/dns.go
+++ b/pkg/bosh/manifest/dns.go
@@ -124,6 +124,13 @@ func (dns *boshDomainNameService) DNSSetting() (corev1.DNSPolicy, *corev1.PodDNS
 	}, nil
 }
 
+var boshDNSDockerImage = ""
+
+// SetupBoshDNSDockerImage initializes the package scoped variable
+func SetupBoshDNSDockerImage(image string) {
+	boshDNSDockerImage = image
+}
+
 // Reconcile see interface
 func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace string, c client.Client, setOwner func(object metav1.Object) error) error {
 	const appName = "bosh-dns"
@@ -172,7 +179,7 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
 						{
 							Name:  "coredns",
 							Args:  []string{"-conf", "/etc/coredns/Corefile"},
-							Image: "coredns/coredns:1.6.3",
+							Image: boshDNSDockerImage,
 							Ports: []corev1.ContainerPort{dnsUDPPort, dnsTCPPort, metricsPort},
 							VolumeMounts: []corev1.VolumeMount{
 								{MountPath: "/etc/coredns", Name: volumeName, ReadOnly: true},
--- a/pkg/bosh/manifest/dns.go
+++ b/pkg/bosh/manifest/dns.go
@@ -12,17 +12,17 @@ import (
 
 	"code.cloudfoundry.org/cf-operator/pkg/kube/util/ctxlog"
 
-	v13 "k8s.io/api/apps/v1"
+	appsv1 "k8s.io/api/apps/v1"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
 
-	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"code.cloudfoundry.org/cf-operator/pkg/kube/util/mutate"
 
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 
 	"code.cloudfoundry.org/cf-operator/pkg/kube/util/names"
 	"github.com/pkg/errors"
@@ -37,11 +37,11 @@ type DomainNameService interface {
 	// HeadlessServiceName constructs the headless service name for the instance group.
 	HeadlessServiceName(instanceGroupName string, deploymentName string) string
 
-	// Configure DNS inside pod
-	ConfigurePod(podSpec *v1.PodSpec)
+	// DNSSetting get the DNS settings for POD
+	DNSSetting() (corev1.DNSPolicy, *corev1.PodDNSConfig)
 
 	// Reconcile DNS stuff
-	Reconcile(ctx context.Context, c client.Client, setOwner func(object v12.Object) error) error
+	Reconcile(ctx context.Context, c client.Client, setOwner func(object metav1.Object) error) error
 }
 
 // Target of domain alias
@@ -123,20 +123,23 @@ func (dns *boshDomainNameService) HeadlessServiceName(instanceGroupName string, 
 	return serviceNames[0]
 }
 
-// ConfigurePod see interface
-func (dns *boshDomainNameService) ConfigurePod(podSpec *v1.PodSpec) {
+// DNSSetting see interface
+func (dns *boshDomainNameService) DNSSetting() (corev1.DNSPolicy, *corev1.PodDNSConfig) {
 	if dns.localDNSIP == "" {
-		panic("BoshDomainNameService: ConfigurePod called before Reconcile")
+		panic("BoshDomainNameService: DNSSetting called before Reconcile")
 	}
-	podSpec.DNSPolicy = v1.DNSNone
-	podSpec.DNSConfig = &v1.PodDNSConfig{
+	return corev1.DNSNone, &corev1.PodDNSConfig{
 		Nameservers: []string{dns.localDNSIP},
 		Searches:    []string{fmt.Sprintf("%s.svc.cluster.local", dns.namespace), "service.cf.internal"},
 	}
 }
 
 // Reconcile see interface
-func (dns *boshDomainNameService) Reconcile(ctx context.Context, c client.Client, setOwner func(object v12.Object) error) error {
+func (dns *boshDomainNameService) Reconcile(ctx context.Context, c client.Client, setOwner func(object metav1.Object) error) error {
+	const appName = "bosh-dns"
+	const volumenName = "bosh-dns-volume"
+	const coreConfigFile = "Corefile"
+	const udpPort = 8053
 
 	rewrites := ""
 	for _, alias := range dns.aliases {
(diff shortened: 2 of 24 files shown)


viovanov:
Thanks @kramerul  - this is great. 👍 
We'll try to get this in as soon as possible.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+				{Name: "dns", Port: 53, Protocol: "UDP", TargetPort: intstr.FromInt(8053)},
+				{Name: "dns-tcp", Port: 53, Protocol: "TCP", TargetPort: intstr.FromInt(8053)},
+				{Name: "metrics", Port: 9153, Protocol: "TCP", TargetPort: intstr.FromInt(9153)},
+			},
+			Selector: map[string]string{"app": "bosh-dns"},
+			Type:     "ClusterIP",
+		},
+	}
+
+	var mode int32 = 420
```
f0rmiga: This can be global.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: v12.ObjectMeta{
+					Labels: map[string]string{"app": "bosh-dns"},
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  "coredns",
+							Args:  []string{"-conf", "/etc/coredns/Corefile"},
+							Image: "coredns/coredns:1.6.3",
```
f0rmiga: This should come from a flag that we can pass when starting the operator. We can default to this value.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+							Ports: []v1.ContainerPort{
+								{ContainerPort: 8053, Name: "dns-udp", Protocol: "UDP"},
+								{ContainerPort: 8053, Name: "dns-tcp", Protocol: "TCP"},
+								{ContainerPort: 9153, Name: "metrics", Protocol: "TCP"},
+							},
+							VolumeMounts: []v1.VolumeMount{
+								{MountPath: "/etc/coredns", Name: "bosh-dns-volume", ReadOnly: true},
```
f0rmiga: Ideally, `bosh-dns-volume`, `bosh-dns` and `Corefile` should be a consts/variables.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+
+	configMap := v1.ConfigMap{
+		ObjectMeta: metadata,
+		Data:       map[string]string{"Corefile": fmt.Sprintf(corefile, rewrites, dns.namespace, dns.rootDNSIP())},
+	}
+	service := v1.Service{
+		ObjectMeta: metadata,
+		Spec: v1.ServiceSpec{
+			Ports: []v1.ServicePort{
+				{Name: "dns", Port: 53, Protocol: "UDP", TargetPort: intstr.FromInt(8053)},
```
f0rmiga: You can reference the TargetPort by name and make it a const/variable that you can reuse below in the pod spec.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+	}
+	for _, obj := range []v12.Object{&configMap, &deployment, &service} {
+		err := setOwner(obj)
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, c, &configMap, configMapMutateFn(&configMap))
```
f0rmiga: [nit]
```
if _, err := ...; err != nil {
    ...
}
```

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		content = []byte{}
+	}
+	return GetNameserverFromResolveConfig(content)
+}
+
+// GetNameserverFromResolveConfig read nameserver from resolve.conf
+func GetNameserverFromResolveConfig(content []byte) string {
+	found := re.FindSubmatch(content)
+	if len(found) == 0 {
+		ctxlog.Errorf(context.TODO(), "No nameserver found in resolve.conf using 1.1.1.1")
```
f0rmiga: I don't think `context.TODO()` is used correctly here. Shouldn't it be `context.Background()`?

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+var _ DomainNameService = &boshDomainNameService{}
+var simple DomainNameService = &simpleDomainNameService{}
+
+// BoshDNSAddOnName name of bosh dns add on
+const BoshDNSAddOnName = "bosh-dns-aliases"
+
+// NewSimpleDomainNameService emulates old behaviour without bosh dns
+func NewSimpleDomainNameService() DomainNameService {
```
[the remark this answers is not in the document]
f0rmiga: Could you also move the `New*` functions to right after the definition of the struct they construct?

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	updated := deployment.DeepCopy()
+	return func() error {
+		deployment.Labels = updated.Labels
+		deployment.Annotations = updated.Annotations
+		deployment.Spec = updated.Spec
+		return nil
+	}
+}
+
+const corefile = `
```
f0rmiga: How about we use `text/template` here to make it more robust?

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(serviceNames) == 0 {
+		return serviceName(instanceGroupName, deploymentName, 63)
+	}
+	return serviceNames[0]
+}
+
+// ConfigurePod see interface
+func (dns *boshDomainNameService) ConfigurePod(podSpec *v1.PodSpec) {
+	if dns.localDNSIP == "" {
+		panic("BoshDomainNameService: ConfigurePod called before Reconcile")
```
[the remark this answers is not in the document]
kramerul: It could happen if someone calls `ConfigurePod` before `Reconcile`. 
Reading the `localIP` inside `ConfigurePod` would require passing the go k8s client through many layers. Therefore, I would like to leave it as it is.
@f0rmiga, do you have any suggestions how to handle this?

pkg/bosh/manifest/dns_test.go
```diff
@@ -0,0 +1,232 @@
...
+        ]
+      },
+      "release": "bosh-dns-aliases"
+    }
+  ],
+  "name": "bosh-dns-aliases"
+}
+`
+
+func loadAddOn(data string) *manifest.AddOn {
```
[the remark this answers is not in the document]
kramerul: I removed the parameter from `loadAddOn` to make it clearer, that the error should never happen.

cmd/internal/bpm_configs.go
```diff
@@ -31,6 +33,16 @@ instance group.
...
 		defer func() {
 			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+				fmt.Fprintf(os.Stderr, "Sleeping ...\n")
+				time.Sleep(debugGracePeriod)
+			}
+		}()
+
+		defer func() {
+			if r := recover(); r != nil {
```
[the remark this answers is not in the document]
kramerul: We got an index out of bounds exception. Therefore, we added this code. It was really helpful to read the logs after the crash. I will simplify the code.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+func (dns *boshDomainNameService) HeadlessServiceName(instanceGroupName string, deploymentName string) string {
+	serviceNames := dns.FindServiceNames(instanceGroupName, deploymentName)
+	if len(serviceNames) == 0 {
+		return serviceName(instanceGroupName, deploymentName, 63)
+	}
+	return serviceNames[0]
+}
+
+// ConfigurePod see interface
+func (dns *boshDomainNameService) ConfigurePod(podSpec *v1.PodSpec) {
```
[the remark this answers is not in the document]
kramerul: I will change the method to 
```go
DnsSetting() (v1.DNSPolicy, *v1.PodDNSConfig)
```

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
+
+	"code.cloudfoundry.org/cf-operator/pkg/kube/util/ctxlog"
+
+	v13 "k8s.io/api/apps/v1"
```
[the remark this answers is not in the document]
kramerul: This was inserted by golang. I will change it.

pkg/bosh/manifest/dns.go
kramerul: OK

pkg/bosh/manifest/dns.go
kramerul: OK

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+
+	v1 "k8s.io/api/core/v1"
+
+	"code.cloudfoundry.org/cf-operator/pkg/kube/util/names"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//DomainNameService abstraction
+type DomainNameService interface {
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: We tried it, but it didn't feel right. It looked like doing polymorphism manually. Do you see the benefit in reducing the mock effort as that large?

pkg/bosh/manifest/dns.go
c0d1ngm0nk3y: Changing it to `text/template` was a bit overkill, but we refactored it, so  that it is hopefully less errorprone.

pkg/kube/controllers/boshdeployment/bpm_reconciler.go
```diff
@@ -114,32 +117,43 @@ func (r *ReconcileBPM) Reconcile(request reconcile.Request) (reconcile.Result, e
...
 
-	// Deploy instance groups
 	instance := &bdv1.BOSHDeployment{}
 	err = r.client.Get(ctx, types.NamespacedName{Namespace: request.Namespace, Name: instanceName}, instance)
 	if err != nil {
 		return reconcile.Result{},
 			log.WithEvent(bpmSecret, "GetBOSHDeployment").Errorf(ctx, "Failed to get BoshDeployment instance '%s': %v", instanceName, err)
 	}
 
+	dns := manifest.DNS(request.Namespace)
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: That is  a good point. We will refactor  the DNSPolicy  used, so that we should be able to use different DNS objects. We hopefully upload another version tomorrow.

pkg/bosh/manifest/dns.go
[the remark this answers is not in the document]
f0rmiga: I don't think you should change the general logic, just remove the whole check with the panic. A good comment on the function should be enough. If you still think it is very important to check, return an error instead of panicking.

pkg/bosh/manifest/dns.go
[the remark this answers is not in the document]
f0rmiga: Oh! haha
Thanks.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+		return nil
+	}
+}
+
+const corefile = `
+.:8053 {
+  health
+  %s
+  rewrite name substring service.cf.internal %s.svc.cluster.local
+  forward . %s
```
f0rmiga: This should be only `forward . /etc/resolv.conf`. The CoreDNS pod should use the upstream DNS set by k8s. In this case, you won't need 2 other functions above that parse `/etc/resolv.conf` on the cf-operator.

kramerul: We tried this, but it was not working out of the box. Perhaps, we should give it a second try.

pkg/kube/controllers/boshdeployment/bpm_reconciler.go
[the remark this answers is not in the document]
c0d1ngm0nk3y: done: https://github.com/cloudfoundry-incubator/cf-operator/pull/598/commits/f7115851d84ea77f0bc2b226df7794d92265eb5e

pkg/bosh/manifest/dns.go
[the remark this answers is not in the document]
c0d1ngm0nk3y: > Could you also move the `New*` functions to right after the definition of the struct they construct?

https://github.com/cloudfoundry-incubator/cf-operator/pull/598/commits/92217ef71dafd6a3aa37004108a93faa8d16f191

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+
+// boshDomainNameService is used to emulate Bosh DNS
+type boshDomainNameService struct {
+	namespace  string
+	aliases    []Alias
+	localDNSIP string
+}
+
+var _ DomainNameService = &boshDomainNameService{}
+var simple DomainNameService = &simpleDomainNameService{}
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: https://github.com/cloudfoundry-incubator/cf-operator/pull/598/commits/c5b92993856d8805c8d8bd27b50f4c99090a2e6b

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// boshDomainNameService is used to emulate Bosh DNS
+type boshDomainNameService struct {
+	namespace  string
+	aliases    []Alias
+	localDNSIP string
+}
+
+var _ DomainNameService = &boshDomainNameService{}
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: Necessary? Indeed not. But the error message is clearer and there are no negative implications, right?

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,323 @@
...
+	return simple
+}
+
+// NewBoshDomainNameService create a new DomainNameService
+func NewBoshDomainNameService(namespace string, addOn *AddOn) (DomainNameService, error) {
+	dns := boshDomainNameService{namespace: namespace}
+	for _, job := range addOn.Jobs {
+		aliases := job.Properties.Properties["aliases"]
+		if aliases != nil {
+			aliasesBytes, err := json.Marshal(aliases)
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: done: https://github.com/cloudfoundry-incubator/cf-operator/pull/598/commits/1f9f25cb690a598d95d53ad7de1f58a03bf94439
FYI: https://github.com/mitchellh/mapstructure was not part of `go.mod` just `go.sum`.

pkg/bosh/manifest/dns.go
c0d1ngm0nk3y: done

c0d1ngm0nk3y:
fyi: I think there are some review comments open and at least one integration test is still missing. Tomorrow is public holiday in Germany, so we will pick this up on Monday.

pkg/bosh/manifest/dns.go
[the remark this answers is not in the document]
f0rmiga: Oh, thanks for pointing it out. We might have used it in the past and dropped later. Thanks for switching.

pkg/bosh/manifest/dns.go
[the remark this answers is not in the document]
f0rmiga: It's all good. I cannot see 100% the result of this in my head. We can always iterate on it in the future if needed.

kramerul:
We added an integration test.

pkg/bosh/manifest/dns.go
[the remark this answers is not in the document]
kramerul: We will return an error

pkg/bosh/manifest/dns.go
kramerul: We are using `forward . /etc/resolv.conf` now.

pkg/bosh/manifest/dns.go
phil9909: Why should it be global?
I'd even write it inline, but it's not possible as `&420` is not valid int literal and we need a pointer type.

pkg/bosh/manifest/dns.go
phil9909: done

pkg/bosh/manifest/dns.go
f0rmiga: It _can_ be global, but not a must. This comment came out only because, in general terms, it just needs a single initialization, and it will never change. In C++, it would make sense to have it declared as `static const` right before its usage. Now that I think again, I shouldn't have pointed it out in the first place.

cmd/internal/bpm_configs.go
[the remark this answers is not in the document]
f0rmiga: I can see it being helpful while debugging code under development (or hunting a bug). For production use, we usually don't ship with debugging code.

pkg/bosh/manifest/dns.go
```diff
@@ -0,0 +1,316 @@
...
+				return nil, errors.Wrapf(err, "Loading aliases from manifest")
+			}
+			dns.Aliases = append(dns.Aliases, a...)
+		}
+	}
+	return &dns, nil
+}
+
+// FindServiceNames see interface
+func (dns *boshDomainNameService) FindServiceNames(instanceGroupName string) []string {
```
f0rmiga: Since we are doing DNS "aliasing", I don't think we should touch the services cf-operator currently creates. Is it possible to only create the extra service for the alias and keep the rest untouched?

kramerul: In my point of view, the new DNS resolution should be superfluous in the near future and should be replaced by the k8s DNS resolution. To work in this direction, it's essential to create the service with the name which corresponds to the DNS entry.
The alias with the manifest-name-prefix only exists to avoid breaking changes for existing BOSH deployments like [scf]((https://github.com/SUSE/scf/).
My vision would be to remove the aliases from the DNS resolution if all the DNS related stuff in the OPS files (e.g. [api.yaml](https://github.com/SUSE/scf/blob/98137edf63d15c2b0d49970172ac3b37d50bc096/deploy/helm/scf/assets/operations/instance_groups/api.yaml#L57) ) is cleaned up.

f0rmiga: I pushed a commit that changes how the rewrites work with the current services cf-operator creates. cf-operator prepends the deployment name because we can have more than one installation per namespace. Since the CoreDNS instance is per deployment, the generated rewrites will resolve the correct service. In the case of SCF, if no `.service.cf.internal` is used, the CoreDNS instance of the deployment will simply forward the request to the upstream server. I linked to this PR the draft PR I'm working on the SCF side.

Issue #596, opened 2019-09-29
link("thing").instances.find {|item| item.bootstrap} - returns nil


drnic:
I have a bosh template with:

```
<%= link("redis").instances.find {|redis| redis.bootstrap }.address %>
```

But `template-render` init container fails as below, suggesting `link("redis").instances.find {|redis| redis.bootstrap }` is returning nil. Is this not supported in cf-operator or any protips for fixing?

```
redis-deployment-redis-v1-1 template-render + cf-operator util template-render
redis-deployment-redis-v1-1 template-render 2019/09/29 23:32:29 failed to render template: rendering failed: (erb):180:in `block in get_binding': undefined method `address' for nil:NilClass (NoMethodError)
redis-deployment-redis-v1-1 template-render 	from /usr/lib64/ruby/gems/2.5.0/gems/bosh-template-2.2.1/lib/bosh/template/evaluation_context.rb:115:in `if_p'
redis-deployment-redis-v1-1 template-render 	from (erb):177:in `get_binding'
redis-deployment-redis-v1-1 template-render 	from /usr/lib64/ruby/2.5.0/erb.rb:885:in `eval'
redis-deployment-redis-v1-1 template-render 	from /usr/lib64/ruby/2.5.0/erb.rb:885:in `result'
redis-deployment-redis-v1-1 template-render 	from /usr/lib64/ruby/gems/2.5.0/gems/bosh-template-2.2.1/lib/bosh/template/test/template.rb:22:in `render'
redis-deployment-redis-v1-1 template-render 	from /tmp/bosh-erb-renderer717623379/bosh_erb_renderer.rb:78:in `<main>'
```

PR #607, opened 2019-10-01
Add support for the instance bootstrap method in BOSH links

qu1queee:
- Add new bootstrap field of the type boolean into JobInstance struct
- Add unit tests for ensuring generated job instances have the proper information
- Only set bootstrap to TRUE, if the instance is the one with index 0
- Extend the redis bosh release asset, to support the rendering of
  `redis.conf.erb`, where https://github.com/cloudfoundry-incubator/cf-operator/issues/596
  was originated.

[#168827598](https://www.pivotaltracker.com/story/show/168827598)


viovanov:
resolves #596

PR #616, opened 2019-10-04
Rename "provider" flag to "operator-webhook-use-service-reference"

aduffeck:


qu1queee:
@aduffeck can you add the flag into https://github.com/cloudfoundry-incubator/cf-operator/blob/b265501c29faf831cbe5b088d37a59d85662c46b/deploy/helm/cf-operator/README.md#configuration ?

PR #628, opened 2019-10-08
Update extendedjob docs

rohitsakala:
[#169006633](https://www.pivotaltracker.com/n/projects/2192232/stories/169006633)


docs/controllers/extendedjob.md
```diff
@@ -103,6 +86,12 @@ This is an auxiliary controller that relies on the Errand Controller output. It
 
 - The only supported output type currently is JSON with a flat structure, i.e. all values being string values.
 
+- The developer should ensure that he/she redirects the JSON output to a file named output.json in /mnt/quarks volume mount at the end of the container script.
```
qu1queee: Can you add at the very end of this sentence something like:

See [doc example](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/examples/extended-job/exjob_output.yaml#L12)

rohitsakala: Done.

PR #635, opened 2019-10-10
enable writeonfailure in the new logging solution

rohitsakala:


go.mod
```diff
@@ -53,3 +53,5 @@ require (
 	sigs.k8s.io/controller-runtime v0.2.2
 	sigs.k8s.io/yaml v1.1.0
 )
+
+go 1.13
```
qu1queee: I think this is not needed.

cmd/internal/persist_output.go
```diff
@@ -22,10 +27,61 @@ into a versionsed secret or kube native secret using flags specified to this com
...
+		po := extendedjob.NewPersistOutputInterface(namespace, podName, clientSet, versionedClientSet, outputVolumePath)
+
+		return po.PersistOutput()
 	},
 }
 
 func init() {
 	utilCmd.AddCommand(persistOutputCmd)
+
+	persistOutputCmd.Flags().StringP("output-volume-path", "", "/mnt/quarks", "path to the volume mount")
```
qu1queee: While we assume that an extendedjob will place STDOUT into `/mnt/quarks`, making this configurable here, will not match the actual path where the ExtendedJob generates the file. Trying to achieve this will be more complex, so I guess it will be better to remove this flag, and stick to the default `/mnt/quarks/ `directory.

pkg/kube/controllers/extendedjob/job_create_secret.go
```diff
@@ -33,106 +32,137 @@ var (
...
+		}
+	}
+	return nil
+}
+
+// ConvertOutputToSecretContainer converts json output file
+// of the specified container into a secret
+func (po *PersistOutputInterface) ConvertOutputToSecretContainer(containerIndex int, container corev1.Container, exjob *ejv1.ExtendedJob, errorContainerChannel chan<- error) {
+
+	filePath := filepath.Join(po.outputFilePathPrefix, container.Name, "output.json")
```
qu1queee: Do you need the `container.Name` to be inside the path?

go.mod
rohitsakala: Removed.

cmd/internal/persist_output.go
rohitsakala: Done. Removed.

pkg/kube/controllers/extendedjob/job_create_secret.go
rohitsakala: Yes, as this code is run the side container and volume mount paths are like /mnt/quarks/c1/output.json, /mnt/quarks/c2/output.json.

cmd/internal/persist_output.go
```diff
@@ -22,10 +27,47 @@ into a versionsed secret or kube native secret using flags specified to this com
 			return errors.Errorf("persist-output command failed. cf-operator-namespace flag is empty.")
 		}
 
-		return extendedjob.PersistOutput(namespace)
+		// hostname of the container is the pod name in kubernetes
+		podName, err := os.Hostname()
```
edwardstudy: Is it safe to get pod name? Could we use downward API and inject env?

```
spec:
  containers:
    - name: test-container
      image: gcr.io/google_containers/busybox
      command: [ "/bin/sh", "-c", "env" ]
      env:
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

cmd/internal/persist_output.go
```diff
@@ -22,10 +27,47 @@ into a versionsed secret or kube native secret using flags specified to this com
...
+		return po.PersistOutput()
 	},
 }
 
 func init() {
 	utilCmd.AddCommand(persistOutputCmd)
 }
+
+// authenticateInCluster authenticates with the in cluster and returns the client
+func authenticateInCluster() (*kubernetes.Clientset, *versioned.Clientset, error) {
```
edwardstudy: Could we reuse the package `"code.cloudfoundry.org/cf-operator/pkg/kube/config" like:

https://github.com/cloudfoundry-incubator/cf-operator/blob/08944f9767e2196e101f74139dd9291bb2cae690/cmd/internal/root.go#L52-L58

cmd/internal/persist_output.go
rohitsakala: The kubernetes docs say that pod name available by hostname. [0]

[0] https://kubernetes.io/docs/concepts/containers/container-environment-variables/#container-information

cmd/internal/persist_output.go
rohitsakala: Done.

Issue #634, opened 2019-10-10
Image Pull Policy should be consistent


kramerul:
Currently, the ImagePullPolicy can be configured in the helm chart for the cf-operator deployment. This property is not propagated to subsequent jobs like dm,bpm or ig. If the kubernetes cluster consists of more than 2 nodes, this can lead to inconsistent images used for different jobs.

To ensure that the same images are used for subsequent jobs, the cf-operator should always use a SHA as reference to the docker image. 
Even if the ImagePullPolicy is set to Always it could happen that different images are used for subsequent jobs on different k8s nodes.

kramerul:
It seems to be very difficult to get the SHA inside a running container image.

manno:
The way we calculate the version from Git should make sure, that a single version always corresponds to a specific git commit.
Since docker images are referenced by that version, I fail to understand how this can lead to inconsistent images. 
Kubernetes should just pull the image with the correct version, if it's missing?

kramerul:
> The way we calculate the version from Git should make sure, that a single version always corresponds to a specific git commit.
> Since docker images are referenced by that version, I fail to understand how this can lead to inconsistent images.
> Kubernetes should just pull the image with the correct version, if it's missing?

This only causes problems in development environments. We had trouble with several strange behaviours of the operator. It took us days to find out, that they were caused by different versions of the image running on the different nodes within our cluster.

For productive environments this is not really required.

PR #639, opened 2019-10-11
Propagate image-pull policy into generated pods

phil9909:
With this change we propagate the helm value `image.pullPolicy` into the generated pods (which use the cf-operator image).
This mitigates #634

### DIFF
--- a/cmd/internal/root.go
+++ b/cmd/internal/root.go
@@ -157,7 +157,7 @@ func init() {
 	pf.StringP("docker-image-org", "o", "cfcontainerization", "Dockerhub organization that provides the operator docker image")
 	pf.StringP("docker-image-repository", "r", "cf-operator", "Dockerhub repository that provides the operator docker image")
 	pf.StringP("docker-image-tag", "t", version.Version, "Tag of the operator docker image")
-	pf.StringP("docker-image-pull-policy", "", string(corev1.PullAlways), "Image pull policy")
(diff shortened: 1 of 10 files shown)


viovanov:
@phil9909  lgtm
we might need to move some of this to https://github.com/cloudfoundry-incubator/quarks-utils
ping @manno

qu1queee:
@phil9909 can you pleas add the new property into this table pls https://github.com/cloudfoundry-incubator/cf-operator/tree/master/deploy/helm/cf-operator#configuration

manno:
> @phil9909 lgtm
> we might need to move some of this to https://github.com/cloudfoundry-incubator/quarks-utils
> ping @manno

If this is needed for the extended job image, too, then we need to find a place for the `docker_image.go` in `quarks-utils`.
I'm not so happy with the singleton, maybe we should move the information about the docker image into a proper struct and share it via the config https://github.com/cloudfoundry-incubator/quarks-utils/blob/master/pkg/config/config.go#L31 ?

kramerul:
> @phil9909 can you pleas add the new property into this table pls https://github.com/cloudfoundry-incubator/cf-operator/tree/master/deploy/helm/cf-operator#configuration

This table describes the existing properties in `values.yaml`. We haven't added a new property to this file. We have only used the existing `images.pullPolicy`.

PR #642, opened 2019-10-11
BOSH DNS improvements

f0rmiga:
- Added health checks for the CoreDNS instances.
- Named the DNS pods and configmaps according to the deployment name, allowing for multiple instances in the same namespace as designed.
- Smalls refactors for clarity.


pkg/bosh/manifest/dns.go
```diff
@@ -144,13 +145,15 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
 		},
 	}
 
-	var mode int32 = 420
+	var corefileMode int32 = 420
```
mook-as: I know this was from the old value, but please say `0o644` instead for Unix permission bits.

Also curious why it's not `corefileMode := 0o644`, but I guess you need to (for some reason) specify the type explicitly? Same with `replicas`? (I know that it can't be `const` because we're taking the address…)

pkg/bosh/manifest/dns.go
```diff
@@ -165,14 +168,41 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
...
+										Port:   intstr.FromInt(8080),
+										Scheme: "HTTP",
+									},
+								},
+								FailureThreshold:    5,
+								PeriodSeconds:       10,
+								SuccessThreshold:    1,
+								TimeoutSeconds:      5,
+								InitialDelaySeconds: 60,
+							},
```
mook-as: These are long enough (vertically in my editor) that it might be worthwhile splitting into separate methods… up to you though.

Does having a separate liveness probe make sense in this case?

pkg/bosh/manifest/dns.go
```diff
@@ -207,12 +237,43 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
...
 
+func (dns *boshDomainNameService) createCorefile(namespace string) string {
+	rewrites := make([]string, 0)
+	for _, alias := range dns.Aliases {
+		for _, target := range alias.Targets {
+			from := alias.Domain
+			if target.Query == "_" {
+				from = strings.Replace(from, "_", target.InstanceGroup, 1)
+			}
+			to := fmt.Sprintf("%s.%s.svc.cluster.local", dns.HeadlessServiceName(target.InstanceGroup), namespace)
```
mook-as: `.cluster.local` isn't actually a constant, unfortunately.  Some clusters have setups where that's incorrect… not sure how to get the correct value though.

pkg/bosh/manifest/dns.go
```diff
@@ -207,12 +237,43 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
 	return nil
 }
 
+func (dns *boshDomainNameService) createCorefile(namespace string) string {
+	rewrites := make([]string, 0)
+	for _, alias := range dns.Aliases {
+		for _, target := range alias.Targets {
+			from := alias.Domain
+			if target.Query == "_" {
+				from = strings.Replace(from, "_", target.InstanceGroup, 1)
```
mook-as: This could use a comment (because the replacement here is weird and BOSH-y)

pkg/bosh/manifest/dns.go
```diff
@@ -207,12 +237,43 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
...
+			}
+			to := fmt.Sprintf("%s.%s.svc.cluster.local", dns.HeadlessServiceName(target.InstanceGroup), namespace)
+			rewrites = append(rewrites, fmt.Sprintf("rewrite name exact %s %s", from, to))
+		}
+	}
+
+	tmpl := template.Must(template.New("Corefile").Parse(`
+.:8053 {
```
mook-as: Could we indent this? (I assuem coredns doesn't care about whitespace _that_ much?)

pkg/bosh/manifest/dns.go
```diff
@@ -207,12 +237,43 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
...
+    errors
+    health
+    {{- range $_, $rewrite := .Rewrites }}
```
mook-as: I thought (for `text/template` / `html/template`) that you don't need the `$_` and it assigns values (instead of indices) by default?

(We can do `range $rewrite := .` just fine too, I think, so we don't _need_ the temporary struct… but that probably doesn't hurt anyway)

pkg/bosh/manifest/dns.go
```diff
@@ -207,12 +237,43 @@ func (dns *boshDomainNameService) Reconcile(ctx context.Context, namespace strin
...
+    {{ $rewrite }}
+    {{- end }}
+    forward . /etc/resolv.conf
+    cache 30
+    loop
+    reload
+    loadbalance
```
mook-as: Should we document why we chose these values?

pkg/bosh/manifest/manifest.go
```diff
@@ -179,7 +179,7 @@ func (m *Manifest) loadDNS() {
 			var err error
 			dns, err := NewBoshDomainNameService(addon, m.Name)
 			if err != nil {
-				ctxlog.Errorf(context.TODO(), "Error loading bosh dns configuration: %s", err.Error())
+				ctxlog.Errorf(context.Background(), "Error loading bosh dns configuration: %v", err)
```
mook-as: Hmm are we on new enough golang to use `%w` yet? :D

(Is `Background` the correct context? Should it not be passed in somewhere?)

pkg/bosh/manifest/dns.go
f0rmiga: Go allows both syntaxes:

```go
mode := int32(420)
var mode int32 = 420
```

f0rmiga: And yeah, 0644 makes much more sense...

pkg/bosh/manifest/dns.go
f0rmiga: I copied the probes from the cluster CoreDNS instance. I prefer leaving it as is.

pkg/bosh/manifest/manifest.go
f0rmiga: Well, context.TODO is definitely not the correct context. There are lots of places that LoadYAML is called that doesn't have a context. I'd assume that it's ok to use it here?

mook-as: It's probably okay, yes, but there might have been better ones (that would have needed to be propagated). As a default, it's okay, but not really not _that_ much better than `TODO`.  At least `TODO` makes it clear there's more thinking required.

f0rmiga: You are right. It makes more sense to actually leave the TODO. I'll revert this.

manno: We previously discussed, that we wan to avoid those. Also we want logging as far up as possible. So maybe, since you can't always pass in one of our contexts with a logger, maybe you can just return the error and handle it in the caller?

pkg/bosh/manifest/dns.go
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/e7b336078c1211fc56d12b195406f129cb0e2f27.

pkg/bosh/manifest/dns.go
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/1c14f1e9d52238e87e789958171835ef0757de8b.

pkg/bosh/manifest/dns.go
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/5c99d7112830100b0fccf42bbb8943671bfe4fab.

pkg/bosh/manifest/dns.go
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/5c99d7112830100b0fccf42bbb8943671bfe4fab.

pkg/bosh/manifest/dns.go
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/fb0ce8852ec24a79116cc76359c720ba6e359b4b.

f0rmiga:
Thanks for pointing it out @manno. I didn't notice that the imports got messed up.

pkg/bosh/manifest/manifest.go
f0rmiga: Fixed in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/bf10a4553e1f9941c8767493434a0b2653fb4fa6.

Indeed it made way more sense to propagate the error. Any failure loading the DNS addon aliases is critical from the deployment perspective and should stop the whole thing.

pkg/bosh/manifest/dns.go
f0rmiga: You are right. The cluster domain is configurable. I'll expose a flag on the cf-operator to allow passing it. It should also be available from the Helm values.

f0rmiga: Done in https://github.com/cloudfoundry-incubator/cf-operator/pull/642/commits/797674cd9d380dde4cbfce4a2b2f9764b93a2dd5.

pkg/bosh/manifest/dns.go
```diff
@@ -21,19 +22,34 @@ import (
 	"code.cloudfoundry.org/quarks-utils/pkg/names"
 )
 
-//DomainNameService abstraction
+var (
+	boshDNSDockerImage = ""
+	clusterDomain      = ""
```
mook-as: Using a global is kind of icky, but :man_shrugging:

f0rmiga: I felt the same but didn't come with a better solution at the time. I just went with the current setup.

PR #647, opened 2019-10-15
Ensure bpm spec.id is properly populated

qu1queee:
We render BPM ERB files in two different places,
where the ID field of the JobInstance struct was not managed
in the same way. This PR fix that, and make sure that this
field will be set by:
`<ig_name>-<index>-<job_name>`

[#169084256](https://www.pivotaltracker.com/story/show/169084256)


qu1queee:
@edwardstudy not with the ERB they mentioned, can you request for changes please.

qu1queee:
@edwardstudy I just added your requested changes.

Issue #645, opened 2019-10-15
bpm job crashes with null pointer


kramerul:
With the following BOSHDeploment ( loggregator_agent without nats properites)

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: nats-manifest
data:
  manifest: |
    ---
    name: nats-manifest
    update:
      serial: false
    releases:
    - name: nats
      version: "26"
      url: docker.io/cfcontainerization
      stemcell:
        os: opensuse-42.3
        version: 30.g9c91e77-30.80-7.0.0_257.gb97ced55
    - name: loggregator-agent
      url: docker.io/cfcontainerization
      version: "3.9"
    instance_groups:
    - name: nats
      jobs:
      - name: nats
        release: nats
        properties:
          nats:
            user: admin
            password: 123456
          quarks:
            ports:
            - name: "nats1"
              protocol: "TCP"
              internal: 4222
            - name: "nats1-routes"
              protocol: TCP
              internal: 4223
    - name: loggregator_agent
      release: loggregator-agent
      update:
        serial: true
        properties:
          quarks:
            ports:
---
apiVersion: fissile.cloudfoundry.org/v1alpha1
kind: BOSHDeployment
metadata:
  name: nats-deployment
spec:
  manifest:
    name: nats-manifest
    type: configmap
```

the bpm job crashes with

```
2019/10/15 15:02:03 recovered in f: runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
runtime/debug.Stack(0xc000499730, 0x14f8540, 0x257d1b0)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9d
code.cloudfoundry.org/cf-operator/cmd/internal.glob..func1.1(0xc000499da8)
	/go/src/code.cloudfoundry.org/cf-operator/cmd/internal/bpm_configs.go:35 +0x141
panic(0x14f8540, 0x257d1b0)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
code.cloudfoundry.org/cf-operator/pkg/bosh/manifest.(*InstanceGroupResolver).BPMConfigs(0xc0004b9b00, 0x16, 0xc000042046, 0xb)
	/go/src/code.cloudfoundry.org/cf-operator/pkg/bosh/manifest/cmd_instance_group_resolver.go:56 +0x108
[...]

PR #649, opened 2019-10-16
Respect update serial in bosh manifest

### LINKED ISSUE #636: Respect update.serial setting of BOSHDeployment
Currently all Instance Groups are starting in parallel. For [scf](https://github.com/SUSE/scf), this will produce a lot of container crashes (e.g. api needs uaa running). This increases the boot times, because every crash doubles the time until the next restart takes place.

The boot time could be reduced by respecting the `update.serial` setting in the Manifest.
We did some experiments with  [scf](https://github.com/SUSE/scf)

|          | Serial Boot   | Parallel Boot |
|--------|----------------|---------------|
| Time |  Between 14 and 16 min | Between 20 and 30 minutes|
| Crashes | No crashes               |  Many crashes |

We implemented this with the help of an additional init container, which was running before `boshPreStartInitContainers` and waits until `nslookup` of the headless service of the previous instance group was not failing. (For details have a look at our [branch based on 0.4.0 ](https://github.com/akhinos/cf-operator/tree/poc-init-order))

c0d1ngm0nk3y:
Fixes #636 

This introduces an initContainer waiting for another instance group to be reachable via service if the `update: serial` is set accordingly in the manifest.

This results in faster startup times since there are fewer crashes and `CrashLoopBackOff`s. Also the startup phase looks a bit smoother.

There are still some crashes left (e.g. one container of `api` crashes if the database is not migrated yet, this is done by another container)

There is a special handling of the eirini instance group inside `manifest.go`, because eirini is patched at the wrong position inside the manifest. We will prepare a PR for scf to fix this (https://github.com/SUSE/scf/pull/2980). Afterward, this special handling can be removed.

We tried to write an e2e test. But we had several problems to run e2e tests with multiple instance groups (#645 ).

### DIFF
--- a/cmd/internal/wait.go
+++ b/cmd/internal/wait.go
@@ -1,0 +1,28 @@
+package cmd
+
+import (
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(waitCmd)
+}
+
+var waitCmd = &cobra.Command{
+	Use:   "wait",
+	Short: "Wait for required service",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		for {
+			if _, err := net.LookupIP(args[0]); err == nil {
+				break
+			}
+			fmt.Printf("Waiting for %s to be reachable\n", args[0])
+			time.Sleep(time.Second * 5)
+		}
+	},
+}
--- a/cmd/internal/wait.go
+++ b/cmd/internal/wait.go
@@ -9,7 +9,7 @@ import (
 )
 
 func init() {
-	rootCmd.AddCommand(waitCmd)
+	utilCmd.AddCommand(waitCmd)
 }
 
 var waitCmd = &cobra.Command{
--- a/cmd/internal/wait.go
+++ b/cmd/internal/wait.go
@@ -5,21 +5,31 @@ import (
 	"net"
 	"time"
 
+	"github.com/spf13/viper"
+
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	utilCmd.AddCommand(waitCmd)
+	waitCmd.Flags().IntP("timeout", "", 30*60, "timeout in seconds after the required service must be available")
+	viper.BindPFlag("timeout", waitCmd.Flags().Lookup("timeout"))
 }
 
+// waitCmd is used to wait for a service (e.g. database), which is required for the calling job (e.g. cloud-controller).
+// This command is used to implement the update.serial flag in the BOSH manifest
 var waitCmd = &cobra.Command{
 	Use:   "wait",
 	Short: "Wait for required service",
 	Args:  cobra.ExactArgs(1),
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
+		expiry := time.Now().Add(time.Duration(viper.GetInt("timeout")) * time.Second)
 		for {
+			if time.Now().After(expiry) {
+				return fmt.Errorf("timeout during waiting for %s to be reachable", args[0])
+			}
 			if _, err := net.LookupIP(args[0]); err == nil {
-				break
+				return nil
 			}
 			fmt.Printf("Waiting for %s to be reachable\n", args[0])
 			time.Sleep(time.Second * 5)
--- a/pkg/bosh/converter/resolver.go
+++ b/pkg/bosh/converter/resolver.go
@@ -106,8 +106,9 @@ func (r *ResolverImpl) WithOpsManifest(ctx context.Context, instance *bdv1.BOSHD
 		}
 	}
 
+	instanceName := instance.GetName()
 	// Reload the manifest after interpolation, and apply implicit variables
-	manifest, err := bdm.LoadYAML(bytes)
+	manifest, err := bdm.LoadYAMLWithName(bytes, &instanceName)
 	if err != nil {
 		return nil, []string{}, errors.Wrapf(err, "Loading yaml failed in interpolation task after applying ops %#v", m)
 	}
@@ -135,6 +136,8 @@ func (r *ResolverImpl) WithOpsManifest(ctx context.Context, instance *bdv1.BOSHD
 	if err != nil {
 		return nil, varSecrets, errors.Wrapf(err, "failed to apply addons")
 	}
+
+	manifest.CalculateRequiredServices()
 
 	return manifest, varSecrets, err
 }
@@ -181,8 +184,9 @@ func (r *ResolverImpl) WithOpsManifestDetailed(ctx context.Context, instance *bd
 		log.Debugf(ctx, "Applied ops file '%s' for deployment '%s'", op.Name, instance.Name)
 	}
 
+	instanceName := instance.GetName()
 	// Reload the manifest after interpolation, and apply implicit variables
-	manifest, err := bdm.LoadYAML(bytes)
+	manifest, err := bdm.LoadYAMLWithName(bytes, &instanceName)
 	if err != nil {
 		return nil, []string{}, errors.Wrapf(err, "Loading yaml failed in interpolation task after applying ops %#v", m)
 	}
@@ -210,6 +214,8 @@ func (r *ResolverImpl) WithOpsManifestDetailed(ctx context.Context, instance *bd
 	if err != nil {
 		return nil, varSecrets, errors.Wrapf(err, "failed to apply addons")
 	}
+
+	manifest.CalculateRequiredServices()
 
 	return manifest, varSecrets, err
 }
--- a/pkg/bosh/converter/resolver.go
+++ b/pkg/bosh/converter/resolver.go
@@ -106,9 +106,8 @@ func (r *ResolverImpl) WithOpsManifest(ctx context.Context, instance *bdv1.BOSHD
 		}
 	}
 
-	instanceName := instance.GetName()
 	// Reload the manifest after interpolation, and apply implicit variables
-	manifest, err := bdm.LoadYAMLWithName(bytes, &instanceName)
+	manifest, err := bdm.LoadYAML(bytes)
 	if err != nil {
 		return nil, []string{}, errors.Wrapf(err, "Loading yaml failed in interpolation task after applying ops %#v", m)
 	}
@@ -184,9 +183,8 @@ func (r *ResolverImpl) WithOpsManifestDetailed(ctx context.Context, instance *bd
 		log.Debugf(ctx, "Applied ops file '%s' for deployment '%s'", op.Name, instance.Name)
 	}
 
-	instanceName := instance.GetName()
 	// Reload the manifest after interpolation, and apply implicit variables
-	manifest, err := bdm.LoadYAMLWithName(bytes, &instanceName)
+	manifest, err := bdm.LoadYAML(bytes)
 	if err != nil {
 		return nil, []string{}, errors.Wrapf(err, "Loading yaml failed in interpolation task after applying ops %#v", m)
 	}
--- a/pkg/bosh/manifest/instance_group.go
+++ b/pkg/bosh/manifest/instance_group.go
@@ -1,6 +1,8 @@
 package manifest
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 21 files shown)


loewenstein:
> There are still some crashes left (e.g. one container of api crashes if the database is not migrated yet, this is done by another container)

I suppose this will be handled in https://github.com/SUSE/scf/issues/2975

c0d1ngm0nk3y:
> > There are still some crashes left (e.g. one container of api crashes if the database is not migrated yet, this is done by another container)
> 
> I suppose this will be handled in [SUSE/scf#2975](https://github.com/SUSE/scf/issues/2975)

I think this is another issue, because this focuses on uaa (I assume).

We have at least 3 types of crashes left
1) One container of `api` crashes because another container did not migrate the database yet
2) `cc-worker` crashes because the database is not migrated yet. But `cc-worker` has no `update: serial`, so I wonder how this works in `BOSH`
3) Even *WITH* the database migrated `api` crashes, but "only" 2 times or so.

jpalmerLinuxFoundation:
Hi Team, 
EasyCLA check indicates all authors are covered by a signed CLA.
image
Current functionality does not update the initial failure comment (CLA Not Signed) when check passes subsequently.  This was done to reduce comments made to PRs, but we're working on a change to update the comment if one exists on the PR.
https://github.com/communitybridge/easycla/issues/269

c0d1ngm0nk3y:
@f0rmiga @manno Could you have a look? Any comments?

kramerul:
I reopened https://github.com/SUSE/scf/pull/2980 as https://github.com/SUSE/kubecf/pull/60

c0d1ngm0nk3y:
It would be great to get this merged for 3 reasons

1) It shortens the startup times - mainly because there are less `CrashLoopBackOff`
2) It brings some order to the startup removes some race conditions - see https://github.com/SUSE/kubecf/issues/3
3) The startup  looks  much smoother without many crashes that are "normal" because some pod is not ready yet.

Is there any downside?

viovanov:
@c0d1ngm0nk3y we haven't been able to review this yet, it will probably happen next week

c0d1ngm0nk3y:
> @c0d1ngm0nk3y we haven't been able to review this yet, it will probably happen next week

It would be great to get some feedback on this to get this hopefully merged soon. The fewer `CrashLoopBackOff`s would be great to have imho.

kramerul:
> Hi,
> Would love to see that, but I'm focusing on #663 and would prefer that one first. However conflicts are minimal.
> 
> Could you squash the commits a little?
> I'm going to ask around in our daily about the manifest name, I don't think we need `LoadYAMLByName` as we enforce the name to be the BDPL resource name.

The enforcement is done at a later point in time (I think after one serialise/deserialise cycle). For this PR, we need the correct name immediately.

pkg/bosh/converter/resolver.go
```diff
@@ -106,8 +106,9 @@ func (r *ResolverImpl) WithOpsManifest(ctx context.Context, instance *bdv1.BOSHD
 		}
 	}
 
+	instanceName := instance.GetName()
```
manno: Took me a while to figure out `instance` is actually a bdpl resource :)

pkg/bosh/manifest/instance_group.go
```diff
@@ -25,22 +27,72 @@ func (instanceGroups InstanceGroups) InstanceGroupByName(name string) (*Instance
...
+	Env                AgentEnv                `json:"env,omitempty"`
+}
+
+// InstanceGroupQuarks represents the quark property of a InstanceGroup
+type InstanceGroupQuarks struct {
+	RequiredService *string `json:"required_service,omitempty"`
+}
+
+// InstanceGroupProperties represents the properties map of a InstanceGroup
+type InstanceGroupProperties struct {
```
manno: Can this be solved like this https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/bosh/manifest/job.go#L63-L66

I kind of remember that we don't need the JSON workaround with `sigs.k8s.io/yaml`.

manno:
We moved this PR upwards in the backlog: https://www.pivotaltracker.com/n/projects/2192232

pkg/bosh/manifest/manifest.go
```diff
@@ -158,6 +158,11 @@ type duplicateYamlValue struct {
 
 // LoadYAML returns a new BOSH deployment manifest from a yaml representation
 func LoadYAML(data []byte) (*Manifest, error) {
+	return LoadYAMLWithName(data, nil)
+}
+
+// LoadYAMLWithName returns a new BOSH deployment manifest from a yaml representation with the given name
+func LoadYAMLWithName(data []byte, name *string) (*Manifest, error) {
```
viovanov: This should be set as soon as possible, and should already be set in the desired manifest.
Do you need it _before_ the desired manifest is a thing?

kramerul: We started without this change. But it turned out that we got the wrong service names (DNS entries) to wait for.

pkg/bosh/manifest/manifest.go
```diff
@@ -192,6 +200,46 @@ func (m *Manifest) loadDNS() error {
...
+		}
+		if ig.Update != nil && ig.Update.Serial != nil {
+			serial = *ig.Update.Serial
+		}
+
+		//FIXME This is required because eirini is inserted as last instance group in https://github.com/SUSE/scf
+		// Without this condition, eirini would wait for all other instance groups. On the other hand, there are
+		// instance groups, which require eirini. Therefore, without this condition there will be a deadlock
+		// This can be removed, if the position of eirini in scf is fixed.
+		if strings.Contains(ig.Name, "eirini") {
```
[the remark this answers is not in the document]
kramerul: I removed this special handling for eirini.

pkg/bosh/manifest/manifest.go
kramerul: You can see it also in the changed test `resolver_test.go`, that the name of the loaded  manifest was not correct. 
It was quite confusing for us, that the name was not correct when the manifest is loaded the first time.

manno: Historically we assumed the bdpl name would be equal to the manifest name. I think we discussed to add a check for that, but must have somehow ended up with the enforcement in `deployment_reconciler.go`.

The behavior/expectation is also documented, at least in once place: https://github.com/cloudfoundry-incubator/cf-operator/blame/master/docs/from_bosh_to_kube.md#L60

That's also the problem with the [resolver_test.go](https://github.com/cloudfoundry-incubator/cf-operator/blob/5c7bb82f371627cf92516d321ab26c4b314b95f3/pkg/bosh/converter/resolver_test.go#L136), the test data is 'wrong'.

I'm not sure we have the best approach here, convention and partial enforcement. Maybe it makes sense to ignore the bdpl name and use the manifest name everywhere. Syncing these two values everywhere in the code seems brittle, as you discovered ;)

However, I'd rather have the  clean up in a proper story so we have a consistent approach to this design issue?

cmd/internal/wait.go
```diff
@@ -0,0 +1,28 @@
...
+func init() {
+	rootCmd.AddCommand(waitCmd)
+}
+
+var waitCmd = &cobra.Command{
+	Use:   "wait",
+	Short: "Wait for required service",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		for {
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: We can add this of course, bur I would like to understand the reasoning/benefit. This would result in crashing the container, right? And this is one thing this `PR` is  addressing. We  could  of course set the default to `0`, but for what do we need it then?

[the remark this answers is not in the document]
f0rmiga: We don't want to avoid it to crash at all costs. I think crashing the init container after a sensible timeout is reached is useful for telling us that something bad happened, instead of waiting forever.

pkg/bosh/manifest/manifest.go
c0d1ngm0nk3y: > How about we always use the bdpl name, and we don't even [de]serialize the name?
> Then you **always** have to provide the name when you call LoadYaml, and you can error if it's empty.

You mean in a separate story/PR, right?

cmd/internal/wait.go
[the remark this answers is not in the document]
c0d1ngm0nk3y: We have added a timeout. But I would argue that a bit more determinism would help the cf-operator a lot. Why crash and just try it again. If it works, errors could be hidden. But the timeout should not hurt for sure.

cmd/internal/wait.go
```diff
@@ -0,0 +1,28 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(waitCmd)
+}
+
+var waitCmd = &cobra.Command{
```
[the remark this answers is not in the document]
c0d1ngm0nk3y: Done

pkg/bosh/manifest/manifest.go
manno: Seems complicated to do, e.g. the `variable-interpolation` command takes a file name and doesn't know the name of the bdpl resource.

cmd/internal/wait.go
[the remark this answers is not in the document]
f0rmiga: I think the main advantage of having the timeout is that it tells us that something went wrong rather than never telling us anything.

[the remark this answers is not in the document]
c0d1ngm0nk3y: Debatable :) If it doesn't start at all, you would notice. I would agree if it would not just  be restarted. I am not a big fan of "eventually consistent". We see it from time to time that the `cf-operator` behaves strange (e.g. not creating `StatefulSets` for 1 hour), but  if we keep the system running long enough it turn out to be up and  running.
Which has of course also a good side to it, but I wonder sometime how  indeterministic the `cf-operator` can be.

c0d1ngm0nk3y:
> `cmd/cf-operator/cf-operator` binary file is added in one of the commits. You will need to interactively rebase the commits and edit the commit that introduced it, otherwise, it will pollute master once we merge.

Good catch. I will have a look.

pkg/bosh/manifest/manifest.go
c0d1ngm0nk3y: Should this block this PR? Or is something else missing? It seems that this is not directly  related to this PR, right?

viovanov: @c0d1ngm0nk3y you added this function to the package, and I don't think it makes sense.
As a developer, I don't know why that package is exposing these 2 functions and what the differences are. We can create another story, but I thought my suggestion would be fast to implement.
If you don't think so, please let me know.

loewenstein: @viovanov quoting @manno 
> Seems complicated to do, e.g. the variable-interpolation command takes a file name and doesn't know the name of the bdpl resource.

It seems not to be straight forward. I'd propose to merge this and create a chore to clean up the use of manifest and deployment names. WDYT?

c0d1ngm0nk3y: @viovanov I agree that a 2nd `LoadYaml` is confusing and for sure not what we want in the end. But I  think the existing `LoadYaml` is kind of buggy  since it returns different outcomes depending on when it is called. But `LoadYaml` is called quite often, so we didn't want to change this without  understanding the consequences in depth. I think it is an existing bug and not directly  related to this PR, but we can have a look. I was just a bit careful because of the "Seems complicated to do" @manno

viovanov: Fair enough - we could merge and schedule something else. This has been open for a while.

viovanov: https://www.pivotaltracker.com/story/show/169575453

c0d1ngm0nk3y: Yes, it's been a while and I am keen of getting this in. So, please let me know when you see something that should be addressed/clarified :)

@manno Could you perhaps clarify which change you would like to see to get this merged?

manno: @c0d1ngm0nk3y @loewenstein 
Please focus this PR on the update serial behavior, changing the `LoadYAML` is another discussion and should be a separate story.

I still think you should remove the new `LoadYAMLWithName` from this PR, if you fix the test data I [pointed to earlier](https://github.com/cloudfoundry-incubator/cf-operator/blob/5c7bb82f371627cf92516d321ab26c4b314b95f3/pkg/bosh/converter/resolver_test.go#L136).

kramerul:
@viovanov,

simply reverting our change in LoadYAML will not work. 
Our feature only works, if you don't change the corresponding unit tests!
The problem is that the service names will be wrong.

manno:
> @viovanov,
> 
> simply reverting our change in LoadYAML will not work.
> Our feature only works, if you don't change the corresponding unit tests!
> The problem is that the service names will be wrong.

I think that's because existing test data didn't care about our 'convention'. Let me try to clean that up.

c0d1ngm0nk3y:
> e60f266



> I think that's because existing test data didn't care about our 'convention'.

@manno By 'convention' you mean that `bdpl name == manifest name`? At  least 3 weeks ago, this was not true for `kubecf`/`scf`. So the code constructed the wrong service names (e.g. `cf-database` instead of `scf-database`.

3 weeks ago, this was definitely  a bug imho. The manifest name was later changed, but too late depending on when you call it. If the 2 should be the same, why synch them at all?

@viovanov Maybe, `kubecf` or `cf-operator` changed in the meantime, so that this is no longer needed. But it was needed when we opened the  PR. Most pods waited for non existing services. So we  should test it with the latest `kubecf`.

viovanov:
@c0d1ngm0nk3y yes, I think I remember that - it was a bug in kubecf.
I think fixing the test is ok for this PR, we can merge, and afterwards we can properly fix this code in https://www.pivotaltracker.com/story/show/169575453

Sorry this took so long 😅

manno:
I tried to fix all the places, so this doesn't turn into a game of whack-a-mole...

https://github.com/cloudfoundry-incubator/cf-operator/commit/8d4635f7f8252b41805ea26ae60d187369cf79c1

c0d1ngm0nk3y:
> I think I remember that - it was a bug in kubecf.

I think you are refering to a different bug. For `kubecf` it is normal that the manifest name (*cf*) is different from the name o the bdpl (*kubecf*), isn't it???

I had my doubts  that this still works after reverting this part and tested installing `kubecf` and nearly no  pod is getting ready because the pods are waiting for a service `cf-xxx` instead of `kubecf-xxx`.
So for `kubecf`, the "convention" that manifest name = bdpl name is still not true. But `kubecf` is the main reason for `cf-operator`, right?

*Summary* `kubecf` cannot be installed with the master of `cf-operator` after this modified PR.
@kramerul , @viovanov , @manno I would suggest we re-add the missing code on Monday to make it work again and figure then out how to continue.

PR #650, opened 2019-10-16
remove persist output command

manno:
[#169094953](https://www.pivotaltracker.com/story/show/169094953)


viovanov:
@manno just the conflict missing

PR #652, opened 2019-10-18
Fix update behavior, e.g. when a secret was rotated

aduffeck:
This pr and the counterparts in `quarks-utils` and `quarks-job` fix several things when the desired manifest changes, e.g. because a secret was rotated:

* the `bpm` secrets are no longer bumped to identical versions
* only affected `ig-resolved` secrets - and thus only affected instance groups - are bumped
* `ig-resolved` secrets now really only contain the relevant part of the manifest for the according instance group as promised by the spec

Rotating the nats password of a deployed scf v3 leads to this state now:

```
kubectl get pods --namespace scf
NAME                            READY   STATUS    RESTARTS   AGE
scf-adapter-v1-0                5/5     Running   0          23m
scf-api-v1-0                    17/17   Running   6          23m
scf-api-v2-0                    16/17   Running   1          14m
scf-bits-v2-0                   7/7     Running   0          14m
scf-bosh-dns-8659459c49-8t6n5   1/1     Running   0          23m
scf-bosh-dns-8659459c49-jtl4z   1/1     Running   0          23m
scf-cc-worker-v1-0              5/5     Running   5          23m
scf-database-v1-0               5/5     Running   0          23m
scf-diego-api-v1-0              6/6     Running   6          23m
scf-doppler-v2-0                11/11   Running   0          14m
scf-eirini-v2-0                 6/6     Running   0          14m
scf-log-api-v2-0                8/8     Running   0          14m
scf-nats-v2-0                   5/5     Running   0          14m
scf-router-v2-0                 6/6     Running   0          14m
scf-scheduler-v1-0              10/10   Running   10         23m
scf-singleton-blobstore-v2-0    7/7     Running   0          14m
scf-uaa-v2-0                    7/7     Running   0          14m
```
Fixes https://github.com/SUSE/kubecf/issues/45

P.S.: The `routing-api-routing-api` container in the `scf-api-v2-0` pod still doesn't come up properly but that seems to be an unrelated issue:
```
{"timestamp":"2019-10-18T13:18:29.180494804Z","level":"error","source":"routing-api","message":"routing-api.locket-lock.failed-to-acquire-lock","data":{"error":"rpc error: code = AlreadyExists desc = lock-collision","lock":{"key":"routing_api_lock","owner":"scf-api-v2-0","type_code":1},"request-uuid":"6b50f96c-0e8a-4b01-58f0-0ffda77cb093","session":"8","ttl_in_seconds":15}}
```

[#169031866](https://www.pivotaltracker.com/story/show/169031866)

Dependent PR's
1. https://github.com/cloudfoundry-incubator/quarks-utils/pull/12
2. https://github.com/cloudfoundry-incubator/quarks-job/pull/21


pkg/kube/util/reference/reconciles.go
```diff
@@ -63,14 +63,14 @@ func GetReconciles(ctx context.Context, client crc.Client, reconcileType Reconci
...
-		if versionedSecret && reconcileType != ReconcileForExtendedStatefulSet {
+		if versionedSecret {
 			keys := make([]string, len(objectReferences))
 			i := 0
 			for k := range objectReferences {
 				keys[i] = k
 				i++
 			}
-			ok := vss.ContainsSecretName(keys, name)
+			ok := vss.ContainsOutdatedSecretName(keys, name)
```
edwardstudy: could we add some comments that we should reconcile if the versioned secret has bumped? Just for readability.

pkg/bosh/manifest/cmd_instance_group_resolver.go
```diff
@@ -69,7 +69,10 @@ func (dg *InstanceGroupResolver) Manifest() (Manifest, error) {
 		return Manifest{}, err
 	}
 
-	return dg.manifest, nil
+	igManifest := dg.manifest
+	igManifest.InstanceGroups = []*InstanceGroup{dg.instanceGroup}
```
edwardstudy: Could we add some comments for this pr purpose?

pkg/kube/util/reference/reconciles.go
aduffeck: done

pkg/bosh/manifest/cmd_instance_group_resolver.go
aduffeck: Not sure what you're looking for exactly. This pr essentially just makes the code behave like specified [in](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/rendering_templates.md#calculation-of-required-properties-for-an-instance-group-and-bpm-info) [the](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/controllers/extendedstatefulset.md#watches-in-sts-controller) [docs](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/controllers/extendedstatefulset.md#automatic-restart-of-containers)

aduffeck: Not sure what you are looking for exactly. This PR essentially makes the code behave like specified [in](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/rendering_templates.md#calculation-of-required-properties-for-an-instance-group-and-bpm-info) [the](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/controllers/extendedstatefulset.md#watches-in-sts-controller) [docs](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/controllers/extendedstatefulset.md#automatic-restart-of-containers).

edwardstudy: Thx. 

So we have expected behavior that these versioned secrets bump especially only affected ig-resolved secrets. Is it right? I hope it does not conflict with our docs.

> - the bpm secrets are no longer bumped to identical versions
> - only affected ig-resolved secrets - and thus only affected instance groups - are bumped

aduffeck: Yes, since `ig-resolved` secrets are supposed to only include the parts of the desired manifest that are relevant to one instance group they should only change if there are actual changes for it. And that's how it's specified imo, to quote the docs:
```
[The ig-resolved yaml] is the "Resolved Instance Group Properties" yaml file. It contains a deployment manifest structure that only has information pertinent to an instance group
```

PR #656, opened 2019-10-21
Bump to latest quarks-job

f0rmiga:
This fixes the issue of passing a specific service account to an extended job.


pkg/bosh/converter/docker_image.go
```diff
@@ -3,28 +3,31 @@ package converter
...
 
-	"code.cloudfoundry.org/quarks-job/pkg/kube/controllers/extendedjob"
+	"code.cloudfoundry.org/quarks-utils/pkg/config"
 	"code.cloudfoundry.org/quarks-utils/pkg/names"
 )
 
 // operatorDockerImage is the location of the operators own docker image
-var operatorDockerImage = ""
+var operatorDockerImage string
+var operatorImagePullPolicy corev1.PullPolicy
```
edwardstudy: I guess we will set invalid value in container.ImagePullPolicy over GetOperatorImagePullPolicy() if `operatorImagePullPolicy`  == "". How about set default value?

f0rmiga: Done in https://github.com/cloudfoundry-incubator/cf-operator/pull/656/commits/624e0cf08357e77bd37ffde20b0b21fc4c8d5794.

PR #663, opened 2019-10-22
Make current master branches of quarks-job and quarks-utils compatible

manno:
This was meant to bring the shared command helper into cf-operator, but incompatibilities in the master branches made this more difficult. It now contains the work from https://github.com/cloudfoundry-incubator/cf-operator/pull/669 and a bug fix to the pull policy.
It also updates the persist-output code to work with quarks-job master, that code will be removed once we require the quarks-job docker image in integration tests.


cmd/internal/root.go
```diff
@@ -73,14 +61,16 @@ var rootCmd = &cobra.Command{
 			corev1.PullPolicy(viper.GetString("docker-image-pull-policy")),
 		)
 		if err != nil {
-			return wrapError(err, "Couldn't parse cf-operator docker image reference.")
+			wrapError(err, "")
```
edwardstudy: Why didn't return the error?

PR #669, opened 2019-10-29
Update cf-operate with the exjob spec change that is to use Job Spec instead of Pod Spec

rohitsakala:
[#169306262](https://www.pivotaltracker.com/n/projects/2192232/stories/169306262)

The following PR should be merged at once with this PR
1. https://github.com/cloudfoundry-incubator/quarks-job/pull/16

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module code.cloudfoundry.org/cf-operator
 
 require (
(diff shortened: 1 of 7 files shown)


rohitsakala:
Closing since this PR is now part of @manno 's big PR [0]

[0] https://github.com/cloudfoundry-incubator/cf-operator/pull/670

PR #671, opened 2019-10-30
Fix PVC handling when scaling

phil9909:
### Motivation

We tried to scale [kubecf](https://github.com/SUSE/kubecf) and noticed the instance groups using persistent disks had problems.

### Changes

To fix this problems, there are multiple changes required:

* Previously the controller of `BOSHDeployment` created `PVC`s manually instead of using the `volumeClaimTemplates` provided by `(Extended)StatefulSet`. It only created one `PVC` per instance group (instead of `n` where `n` is the number of replicas).
* The `volume-management` pods were using the actual workload, which despite from being slow caused problems (the bosh-rendering fails due to "az-index is negative. -1"). I changed it to use a dummy workload.
* The `volume-management` pods sometimes didn't become ready as the actual workload was faster mounting the volume. The solution was to terminate the pod earlier, i.e. when the pod is "initialized" which also means the `PVC`s were already created.
* The webhook which modifies the `volumeMounts` and `volumes` used a different logic to generate the name of the `PVC`.


phil9909:
@rohitsakala Could you look into this please?

cmd/internal/sleep.go
```diff
@@ -0,0 +1,32 @@
...
+
+This is used as a dummy command for the volume-management-pods.
+
+`,
+	RunE: func(cmd *cobra.Command, args []string) (err error) {
+		return Sleep()
+	},
+}
+
+// Sleep will sleep indefinitely
```
manno: It sleeps 24 hours.

pkg/kube/controllers/extendedstatefulset/pod_mutator.go
```diff
@@ -160,7 +160,7 @@ func (m *PodMutator) addVolumeSpec(pod *corev1.Pod, volumeClaimTemplatesMap map[
 			_, foundVolumeClaimTemplate := volumeClaimTemplatesMap[volumeMount.Name]
 			if foundVolumeClaimTemplate {
 				podOrdinal := names.OrdinalFromPodName(pod.GetName())
-				persistentVolumeClaim := names.Sanitize(fmt.Sprintf("%s-%s-%s-%d", volumeMount.Name, "volume-management", extendedStatefulSetName, podOrdinal))
+				persistentVolumeClaim := fmt.Sprintf("%s-%s-%s-%d", volumeMount.Name, "volume-management", extendedStatefulSetName, podOrdinal)
```
manno: This doesn't need [Sanitize](https://github.com/cloudfoundry-incubator/quarks-utils/blob/master/pkg/names/names.go#L108)? It looks like it could go over the allowed character limit?

pkg/kube/controllers/extendedstatefulset/volumes.go
```diff
@@ -154,11 +157,36 @@ func (r *ReconcileExtendedStatefulSet) generateVolumeManagementSingleStatefulSet
...
+
+	statefulSet.Spec.Template.Spec.InitContainers = []corev1.Container{}
+	statefulSet.Spec.Template.Spec.Containers = []corev1.Container{{
+		Name:            "volume-management",
+		VolumeMounts:    volumeMounts,
+		Image:           converter.GetOperatorDockerImage(),
+		ImagePullPolicy: converter.GetOperatorImagePullPolicy(),
+		Args: []string{
+			"util",
+			"sleep",
```
manno: Is it  really necessary to add `sleep` as a command? Wouldn't it be sufficient to shell out and call `bash -c 'sleep N'` or even `read`? It's already a shell-out, so why not reuse something we know we have on our image?

cmd/internal/sleep.go
phil9909: Inside an endless loop :wink:

pkg/kube/controllers/extendedstatefulset/pod_mutator.go
phil9909: > Kubernetes resources can have names up to 253 characters long.
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

I think the lower limit is only valid for pods.

Still it is possible that we hit that higher limit. But at this part of the code it's already to late to sanitize. We should check what happens if we create a `StatefulSet` with a `volumeClaimTemplate.name` length 253  as the StatefulSet controller will automatically append `-1`, `-2`, etc. which we also do here. So we can not use our own `Sanitize` logic, we have to do the exact same thing.

pkg/kube/controllers/extendedstatefulset/volumes.go
phil9909: We thought this would be the pattern to use, but we can also call `bash` or `ruby` if you prefer.
Do you want us to change it?

cmd/internal/sleep.go
manno: I see the comment was on the function :)

pkg/kube/controllers/extendedstatefulset/pod_mutator.go
manno: Ah we had a similar issue with job names, where a digest was appended to the name when using the name for the job's pod.

https://github.com/cloudfoundry-incubator/quarks-utils/blob/master/pkg/names/names.go#L137-L139

Not sure if anything in kubecf will ever reach that higher limit, but to be safe we should allow for `-123`, if someone has a lot of instances maybe?

pkg/kube/controllers/extendedstatefulset/volumes.go
manno: Yes, I would prefer it. We do control the whole docker image anyways. And if running standalone, bash should be available. Running the operator on Windows is not a thing.

phil9909: I used `ruby` which we need anyway for running the bosh templating logic (I guess).

pkg/kube/controllers/extendedstatefulset/pod_mutator.go
phil9909: I just tried to create a `StatefulSet` with `volumeClaimTemplate` with a very long (253 chars) name. No pod is ever created and `describe StatefulSet` shows error events. So we don't need to consider this case in the `PodMutator`, but handle it before creating the `volume-management`-`StatefulSet`.

I can fix this, but would prefer to do it in a separate PR. ok?

```yaml
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  selector:
    matchLabels:
      app: nginx
  serviceName: "nginx"
  template:
    metadata:
      labels:
        app: nginx
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - name: nginx
        image: k8s.gcr.io/nginx-slim:0.8
        ports:
        - containerPort: 80
          name: web
        volumeMounts:
        - name: wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww
          mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
  - metadata:
      name: wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww
    spec:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "default"
      resources:
        requests:
          storage: 1Gi
```

manno: yes, sounds good

phil9909: Lets first see what the outcome of [this](https://www.pivotaltracker.com/n/projects/2192232/stories/169531360) story is. If we can get rid of the versioned stateful sets we also won't need the volumeManagement anymore.

PR #678, opened 2019-10-31
Make resource requests configurable

jkbschmid:
To avoid overloading a k8s node which prevents the Bosh deployment from being successfully deployed, it should be possible to set resource request for k8s containers. 
Since k8s containers are created from bpm processes, we have added a corresponding configuration option `manifest.InstanceGroups[0].Jobs[0].Properties.Quarks.BPM.Processes[0].Requests`.

Since the variable templating uses reflection and k8s API structs have private fields, we also had to make a small change there.


c0d1ngm0nk3y:
> Could we add this limit to bpm.resources: https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/from_bosh_to_kube.md#Resources

We added a separate section [Misc](https://github.com/akhinos/cf-operator/blob/configure-requests/docs/from_bosh_to_kube.md#Misc). We haven't added it to `Resources` because it is not available in Bosh itself, that might be misleading.

pkg/bosh/manifest/cmd_instance_group_resolver_test.go
```diff
@@ -170,8 +170,38 @@ var _ = Describe("InstanceGroupResolver", func() {
...
+					ig = "log-api"
+				})
+
+				It("raises an error for bpm process without executable", func() {
+					_, err := dg.BPMConfigs()
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring("invalid BPM process"))
+				})
+			})
+		})
```
manno: nit-pick: missing a newline

pkg/bosh/converter/container_factory_test.go
```diff
@@ -501,6 +525,7 @@ var _ = Describe("ContainerFactory", func() {
 				Expect(len(containers)).To(Equal(1))
 			})
 		})
+
```
manno: nit-pick: superfluous newline

PR #683, opened 2019-11-05
Make helm chart values available to quarks-job

manno:
Moves several helm values to `global:` so they can be used by the quarks-job sub helm chart.

Also adds defaults for the separate quarks-job service account.

This depends on https://github.com/cloudfoundry-incubator/quarks-job/pull/17

[#169095052](https://www.pivotaltracker.com/story/show/169095052)


manno:
Cannot bump quarks-job helm chart before quarks-job PR is merged, since travis only publishes for master

PR #686, opened 2019-11-06
Fixed container-run when exit 1

f0rmiga:


container-run/pkg/containerrun/containerrun.go
```diff
@@ -269,8 +269,12 @@ func (p *ContainerProcess) Signal(sig os.Signal) error {
...
 // Wait waits for the process.
 func (p *ContainerProcess) Wait() error {
-	if _, err := p.process.Wait(); err != nil {
-		return fmt.Errorf("failed to wait for process: %v", err)
+	state, err := p.process.Wait()
+	if err != nil {
+		return fmt.Errorf("failed to run process: %v", err)
+	} else if !state.Success() {
+		err := &exec.ExitError{ProcessState: state}
+		return fmt.Errorf("failed to run process: %v", err)
```
mook-as: It might be nice to have different error messages between "could not run" and "ran, but the process returned an error".

container-run/pkg/containerrun/containerrun_test.go
```diff
@@ -651,6 +650,23 @@ var _ = Describe("ContainerRunner", func() {
 			Expect(p).To(BeNil())
 		})
 
+		It("fails when the command exits 1", func() {
+			cr := NewContainerRunner()
+			cmd := Command{
+				Name: "bash",
+				Arg:  []string{"-c", "exit 1"},
```
mook-as: Any reason to not do this like `succeeds when…` but with an unsucessfull `ProcessState` ?

Or at least do it by setting the command to `/bin/false` or something… (I assume none of this is expected to work on Windows).

container-run/pkg/containerrun/containerrun.go
f0rmiga: It is handled by the error being propagated.

container-run/pkg/containerrun/containerrun_test.go
f0rmiga: They are in different contexts. Here we are testing `Run`, not `Wait`. The scope here is broader. As for `/bin/false` or `exit 1`, they don't really make any practical difference, right?

PR #688, opened 2019-11-06
Update quarks-job dependency to a reviewed version

manno:
This was missing from #683 

[#169095052](https://www.pivotaltracker.com/story/show/169095052)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Temporary Build Files
 /cf-operator
-binaries
+/binaries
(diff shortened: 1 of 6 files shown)


manno:
> @manno: Ah, this was why this morning's combination of operator and KubeCF resulted in an operator waiting but not doing anything for me? And, I now need to prepend `global.` to the value?

Yes, https://cloudfoundry.slack.com/archives/C1BQKKNP4/p1572966288023000

PR #692, opened 2019-11-07
Introduce StatefulSet Reconciler to implement BOSH canary feature. Th…

kramerul:
Introduce StatefulSet Reconciler to implement BOSH canary feature. 
The version handling for stateful sets is removed.

[#169531360](https://www.pivotaltracker.com/n/projects/2192232/stories/169531360)


pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
```diff
@@ -228,44 +226,31 @@ func (r *ReconcileQuarksStatefulSet) createStatefulSet(ctx context.Context, qSta
 		return errors.Wrapf(err, "could not set owner for StatefulSet '%s' to QuarksStatefulSet '%s' in namespace '%s'", statefulSet.Name, qStatefulSet.Name, qStatefulSet.Namespace)
 	}
 
-	// Create the StatefulSet
-	if err := r.client.Create(ctx, statefulSet); err != nil {
-		return errors.Wrapf(err, "could not create StatefulSet '%s' for QuarksStatefulSet '%s' in namespace '%s'", statefulSet.Name, qStatefulSet.Name, qStatefulSet.Namespace)
+	statefulSetMutateFn := func(sfs *v1beta2.StatefulSet) controllerutil.MutateFn {
```
edwardstudy: We had a common package to handle mutates

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strings"
 	"time"
 
+	"code.cloudfoundry.org/cf-operator/pkg/kube/controllers/statefulset"
```
edwardstudy: it's better to reorder dependencies

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
```diff
@@ -108,28 +110,18 @@ func (r *ReconcileQuarksStatefulSet) Reconcile(request reconcile.Request) (recon
...
 	// Get the current StatefulSet.
-	currentStatefulSet, currentVersion, err := GetMaxStatefulSetVersion(ctx, r.client, qStatefulSet)
+	_, currentVersion, err := GetMaxStatefulSetVersion(ctx, r.client, qStatefulSet)
 	if err != nil {
 		return reconcile.Result{}, ctxlog.WithEvent(qStatefulSet, "StatefulSetNotFound").Error(ctx, "Could not retrieve latest StatefulSet owned by QuarksStatefulSet '", request.NamespacedName, "': ", err)
 	}
 
 	// Calculate the desired statefulSets
-	desiredStatefulSets, desiredVersion, err := r.calculateDesiredStatefulSets(qStatefulSet, currentVersion)
+	desiredStatefulSets, _, err := r.calculateDesiredStatefulSets(qStatefulSet, currentVersion)
```
phil9909: Do we still need the currentVersion here

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
```diff
@@ -108,28 +110,18 @@ func (r *ReconcileQuarksStatefulSet) Reconcile(request reconcile.Request) (recon
 	}
 
 	// Get the current StatefulSet.
-	currentStatefulSet, currentVersion, err := GetMaxStatefulSetVersion(ctx, r.client, qStatefulSet)
+	_, currentVersion, err := GetMaxStatefulSetVersion(ctx, r.client, qStatefulSet)
```
phil9909: can we delete this?

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
manno: https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/util/mutate/mutate.go

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
```diff
@@ -281,36 +254,39 @@ func (r *ReconcileQuarksStatefulSet) generateSingleStatefulSet(qStatefulSet *qst
...
 		MatchLabels: labels,
 	}
-	statefulSet.SetAnnotations(annotations)
+	statefulSet.SetAnnotations(mergeMaps(statefulSet.GetAnnotations(), annotations))
 
+	r.injectContainerEnv(&statefulSet.Spec.Template.Spec, zoneIndex, zoneName, qStatefulSet.Spec.Template.Spec.Replicas)
 	return statefulSet, nil
 }
 
+func mergeMaps(maps ...map[string]string) map[string]string {
```
phil9909: Could be replaced by "deepMerge" or so

manno: https://godoc.org/github.com/imdario/mergo#MergeWithOverwrite

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
c0d1ngm0nk3y: We "only" need them for the annotation. It is not needed, but quite nice to have because it is an easy way to see if the stateful set has been updated. I refactored it a bit to be not that confusing.

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
c0d1ngm0nk3y: I moved it

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
c0d1ngm0nk3y: done

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
c0d1ngm0nk3y: We still need the version for the annotation. I have moved it to `calculateDesiredStatefulSets` to be less confusing.

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go
c0d1ngm0nk3y: > "for value type attributes on the dst struct"

@manno The `MergeWithOverwrite` can only be used for structs, right?

manno: @c0d1ngm0nk3y I don't think I ever used it myself, just saw it in the code once and the web page says:

> Mergo: merging Go structs and maps since 2013.

c0d1ngm0nk3y: Strange, I got a reflection error that it is a map value. I will have another look tomorrow.

kramerul: We decided to stay with the old implementation for the following reasons:
* It's simpler, because no reflection is needed
* No error handling required
* Would require additional glue code, because the second map is merged into the first

kramerul:
Hi @viovanov, hi @loewenstein,
the PR is now ready for merge. We did the following stuff:
* Changed the semantic and name of `generation` to `initialRollout`
* Fixed all tests
* Removed the `currentVersion`
* Stick to our implementation of merging maps (see reasons above)
* Moved `mergeMaps`, `minInt64` and `maxInt64` to the `util` package and renamed `mergeMaps` to `UnionMaps`
* Added `-ms` to the annotations `canary-watch-time` and `update-watch-time`
* Set `RequeueAfter` in state `Rollout` to avoid getting stuck in state `Rollout`

Regards Ulrich

pkg/bosh/converter/bpm_resources.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strconv"
 
+	"code.cloudfoundry.org/cf-operator/pkg/kube/controllers/statefulset"
```
manno: just nitpicking: we group imports by: core, github, k8s, quarks

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler_test.go
```diff
@@ -130,12 +130,52 @@ var _ = Describe("ReconcileQuarksStatefulSet", func() {
...
+			It("sets no RollingUpdate even if replica=1", func() {
+				result, err := reconciler.Reconcile(request)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(result).To(Equal(reconcile.Result{}))
+
+				ss := &v1beta2.StatefulSet{}
+				err = client.Get(context.Background(), types.NamespacedName{Name: "foo", Namespace: "default"}, ss)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			Context("with multiple replicas", func() {
```
manno: nit: missing newline

pkg/kube/controllers/statefulset/statefulset_rollout_util_test.go
```diff
@@ -0,0 +1,198 @@
...
+					}
+				}
+			}
+			return apierrors.NewNotFound(schema.GroupResource{}, nn.Name)
+		})
+
+	})
+
+	Context("statefulset is updated", func() {
+
```
manno: nit: one newline too many

pkg/kube/controllers/statefulset/statefulset_rollout_util_test.go
```diff
@@ -0,0 +1,198 @@
...
+		It("ready pod is not deleted", func() {
+			err := statefulset.CleanupNonReadyPod(ctx, client, statefulSet, 0)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(client.GetCallCount()).To(Equal(1))
+			Expect(client.DeleteCallCount()).To(Equal(0))
+		})
+		It("pending pod is deleted", func() {
```
manno: nit: missing newline

manno:
Squash the 'Fix ... test' commits, please.

integration/environment/instance_group.go
```diff
@@ -10,14 +11,18 @@ import (
 
 // WaitForInstanceGroup blocks until all selected pods of the instance group are running. It fails after the timeout.
 func (m *Machine) WaitForInstanceGroup(namespace string, deployment string, igName string, version string, count int) error {
+	return m.WaitForInstanceGroupMultiVersions(namespace, deployment, igName, count, version)
+}
+
+// WaitForInstanceGroupMultiVersions blocks until all selected pods of the instance group are running. It fails after the timeout.
+func (m *Machine) WaitForInstanceGroupMultiVersions(namespace string, deployment string, igName string, count int, versions ...string) error {
```
viovanov: Should this be `WaitForInstanceGroupGenerationAllReplicas` (based on the comment)?
Also, are we calling it `Version` or `Generation`?

pkg/kube/controllers/boshdeployment/validating_webhook.go
```diff
@@ -6,6 +6,10 @@ import (
 	"strings"
 	"time"
 
+	"code.cloudfoundry.org/cf-operator/pkg/kube/controllers/statefulset"
+
```
viovanov: remove extra newline please

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
```diff
@@ -0,0 +1,68 @@
+package statefulset
```
viovanov: This should be in the `quarksstatefulset` package

pkg/kube/controllers/statefulset/statefulset_emulation_test.go
```diff
@@ -0,0 +1,211 @@
+package statefulset_test
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
```
viovanov: please remove extra newlines

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+
```
viovanov: remove newline please
we also group imports in 
- standard lib
- third party
- project

pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
```diff
@@ -0,0 +1,225 @@
+package statefulset
```
viovanov: should live in the `quarksstatefulset` package

pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
```diff
@@ -0,0 +1,225 @@
...
+
+	statefulSet := v1beta2.StatefulSet{}
+
+	err := r.client.Get(ctx, request.NamespacedName, &statefulSet)
+	if err != nil {
+		ctxlog.Debug(ctx, "StatefulSet not found ", request.NamespacedName)
+		return reconcile.Result{}, err
+	}
+	var result reconcile.Result
+	var status = statefulSet.Annotations[AnnotationCanaryRollout]
```
viovanov: status shouldn't live in annotations, it should live in status object of the CRD
https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/apis/quarksstatefulset/v1alpha1/types.go#L65

pkg/kube/controllers/statefulset/statefulset_rollout_util.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strconv"
+	"time"
+
+	"code.cloudfoundry.org/cf-operator/pkg/kube/util"
+
```
viovanov: newlines and grouping please

pkg/kube/controllers/statefulset/statefulset_rollout_util_test.go
```diff
@@ -0,0 +1,198 @@
+package statefulset_test
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/types"
+
```
viovanov: newlines and grouping please

pkg/kube/controllers/statefulset/statefulset_webhook.go
```diff
@@ -0,0 +1,128 @@
+package statefulset
```
viovanov: should live in the `quarksstatefulset` package

pkg/kube/controllers/statefulset/statefulset_rollout_util.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// ConfigureStatefulSetForRollout configures a stateful set for canarying and rollout
+func ConfigureStatefulSetForRollout(statefulSet *v1beta2.StatefulSet) {
+	statefulSet.Spec.UpdateStrategy.Type = v1beta2.RollingUpdateStatefulSetStrategyType
+	//the canary rollout is for now directly started, the might move to a webhook instead
+	statefulSet.Spec.UpdateStrategy.RollingUpdate = &v1beta2.RollingUpdateStatefulSetStrategy{
+		Partition: pointers.Int32(util.MaxInt32(util.MinInt32(*statefulSet.Spec.Replicas-1, statefulSet.Status.Replicas), 0)),
+	}
+	statefulSet.Annotations[AnnotationCanaryRollout] = rolloutStatePending
```
viovanov: should not store state here - state should be set on the qsts

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Adding StatefulSet rollout controller to manager failed.")
+	}
+
+	// Trigger when annotation is set
+	statefulSetPredicates := predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+		UpdateFunc:  CheckUpdate,
```
manno: We use 'predicate logging' on all controllers to generate events and debug information.

Something like this: 
https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/controllers/boshdeployment/deployment_controller.go#L62-L65

pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
```diff
@@ -0,0 +1,225 @@
...
+type ReconcileStatefulSetRollout struct {
+	ctx    context.Context
+	client crc.Client
+	scheme *runtime.Scheme
+	config *config.Config
+}
+
+// Reconcile cleans up old versions and volumeManagement statefulSet of the ExtendedStatefulSet
+func (r *ReconcileStatefulSetRollout) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+
```
manno: (nit: newline)

This reconciler doesn't use the 'meltdown' package to protect against too many reconciles. (https://github.com/cloudfoundry-incubator/cf-operator/blob/master/pkg/kube/controllers/boshdeployment/deployment_reconciler.go#L89-L92)

There are two ways to use meltdown, with annotations or with a status field.

I actually wonder how this worked without it, maybe our test cases are not complicated enough to trigger enough reconcile events?

The main problem I see with the state machine approach is that it relies on an annotation, while the actual cluster state might be different, since the annotation has not been updated yet.
(see "level based api" in https://book-v1.book.kubebuilder.io/basics/what_is_a_controller.html)

Should work though, if the reconciler doesn't take too long?

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
loewenstein: @viovanov Actually the controller is usable standalone.

pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
loewenstein: @viovanov the controller is a _decorator_ for StatefulSets, so we should have the status on the StatefulSet. Reusing the StatefulSet status is dangerous though, hence the decision to go for annotation.

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
c0d1ngm0nk3y: This conflicts with another comment https://github.com/cloudfoundry-incubator/cf-operator/pull/692#discussion_r352036555 by @manno 
which says:
* standard lib
* third party
* k8s
* project

What is correct?

pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler_test.go
c0d1ngm0nk3y: done

pkg/kube/controllers/statefulset/statefulset_rollout_util_test.go
c0d1ngm0nk3y: done

pkg/kube/controllers/statefulset/statefulset_rollout_util_test.go
c0d1ngm0nk3y: done

pkg/bosh/converter/bpm_resources.go
c0d1ngm0nk3y: Shouldn't the `lint` not check things like this?  Also, how to ensure this? We have tried to configure `goimport` in that way, but without success.
I think having it part of the `lint` would be quite important to get early feedback and not have to check x files at the end.

integration/environment/instance_group.go
c0d1ngm0nk3y: We renamed the method.

With regards to `version` vs. `generation` - we would like to stay with `version` because the `generation` used in kubernetes is not exactly the same. Especially changing the `partition` will increase the `generation`.

pkg/kube/controllers/boshdeployment/validating_webhook.go
c0d1ngm0nk3y: We went through all imports.

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
c0d1ngm0nk3y: What would be the benefit of having it in `quarksstatefulset`? I would also prefer to keep it independent if possible.

pkg/kube/controllers/statefulset/statefulset_emulation_test.go
c0d1ngm0nk3y: done

pkg/kube/controllers/statefulset/statefulset_rollout_util.go
c0d1ngm0nk3y: done

pkg/kube/controllers/statefulset/statefulset_rollout_util_test.go
c0d1ngm0nk3y: done

pkg/kube/controllers/statefulset/statefulset_webhook.go
c0d1ngm0nk3y: see above

pkg/kube/controllers/statefulset/statefulset_rollout_util.go
c0d1ngm0nk3y: see above

pkg/kube/controllers/statefulset/statefulset_rollout_controller.go
c0d1ngm0nk3y: We have added it...

c0d1ngm0nk3y:
> Is that sufficient for the new canary feature, i.e. are existing tests covering it?

You are right. We have added another test to check the canarying.

manno:
> > Squash the 'Fix ... test' commits, please.
> 
> Why exactly? Should this not be the job of whoever does the merging? There is an option on github, isn't there? Otherewise, how do the reviewer see what has changed since the last review?

I see. Those two stuck out, because they seemed 'technical'.
Regarding reviewing, I proposed this after seeing all the commits were recent and that you already did force push:
> kramerul force-pushed the akhinos:stateful-set branch from 3a8776d to 7180ca8 yesterday

No worries, I don't think anyone was looking at individual commits.

Not sure if we want to squash all commits, which is what Github offers.

pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
viovanov: Having it be independent from the qsts crd is not a requirement, but we can see how this goes.
I don't like not having the state in the qsts, but we can read it from the annotation.

viovanov:
Thanks @c0d1ngm0nk3y @phil9909 @kramerul @loewenstein - I'm ok with the PR.
Can you please rebase and also address @manno's comment about using meldown?

pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
jkbschmid: Done.

jkbschmid:
@viovanov We rebased and added meltdown to our reconciler.

PR #695, opened 2019-11-11
Decrease RSA key size to 2048

kramerul:
According to [Wikipedia](https://en.wikipedia.org/wiki/Key_size#Asymmetric_algorithm_key_lengths):

> As of 2003 RSA Security claims that 1024-bit RSA keys are equivalent in strength to 80-bit 
> symmetric keys, 2048-bit RSA keys to 112-bit symmetric keys and 3072-bit RSA keys to 128-bit 
> symmetric keys.[14] RSA claims that 1024-bit keys are likely to become crackable some time 
> between 2006 and 2010 and that 2048-bit keys are sufficient until 2030.[14] The NIST 
> recommends 2048-bit keys for RSA. 

After 2030, it is questionable that RSA keys are strong enough for quantum computers. Therefore, switching to elliptical curves would be more appropriate than using 4096 bit RSA keys.

This decreases the amount of CPU during installation of kubecf from 300 seconds to less than 30 seconds.


pkg/credsgen/in_memory_generator/in_memory_generator.go
```diff
@@ -16,5 +16,5 @@ type InMemoryGenerator struct {
 
 // NewInMemoryGenerator creates a default InMemoryGenerator
 func NewInMemoryGenerator(log *zap.SugaredLogger) *InMemoryGenerator {
-	return &InMemoryGenerator{Bits: 4096, Expiry: 365, Algorithm: "rsa", log: log}
+	return &InMemoryGenerator{Bits: 2048, Expiry: 365, Algorithm: "rsa", log: log}
```
manno: There is a unit test for that: 

```
/tmp/build/a94a8fe5/src/code.cloudfoundry.org/cf-operator/pkg/credsgen/in_memory_generator/in_memory_generator_test.go:30


  Expected

      <int>: 2048

  to equal

      <int>: 4096
```

phil9909: done

PR #698, opened 2019-11-12
Add optional image pull secret to service account

HeavyWombat:
Add service account image pull secret to cf-operator [#169646805]

In case the `cf-operator` Docker image needs to be stored in a private
registry, the service account needs to know which image pull secret to
use and the respective image pull secret needs to be there.

Add condition to service account deployment YAML to add an image pull
secret reference (imagePullSecrets).

Use aforementioned condition to render a deployment YAML for a secret,
that matches the service account setting.

Add empty credentials entry in `values.yaml` as the default, which means
that no credentials are required.

Usage example:
```
 helm install \
  --namespace cfo \
  --name cf-operator \
  --set global.image.credentials.servername=private.registry.io \
  --set global.image.credentials.username=token \
  --set global.image.credentials.password=$(vault supersecret) \
  https://foobar.com/cf-operator-v0.47.11.tgz
```

### DIFF
--- a/deploy/helm/cf-operator/README.md
+++ b/deploy/helm/cf-operator/README.md
@@ -56,20 +56,21 @@ helm delete cf-operator --purge
 
 ## Configuration
 
-| Parameter                                         | Description                                                                          | Default                                        |
-| ------------------------------------------------- | ------------------------------------------------------------------------------------ | ---------------------------------------------- |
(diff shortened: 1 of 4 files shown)


qu1queee:
Reviewing..

HeavyWombat:
@qu1queee Updated the documentation to include a reference to the new setting.

PR #708, opened 2019-11-15
Propagate canary_watch_time to StatefulSet

jkbschmid:
Needed for https://github.com/cloudfoundry-incubator/cf-operator/pull/692.

To do canary rollouts like Bosh does, we should support the `canary_watch_time` property.
We need to propagate this property from the BoshManifest to the StatefulSet, so the rollout controller can access it.

I also added a validation to the ValidatingWebhook so that we notice early if the property is misconfigured.


jkbschmid:
Will be part of https://github.com/cloudfoundry-incubator/cf-operator/pull/692

PR #714, opened 2019-11-19
Add missing steps to run integration tests on minikube.

jkbschmid:
I (successfully) tried to get the integration test to run on minikube and wanted to documented the missing steps I had to take.


docs/testing.md
```diff
@@ -139,37 +139,53 @@ The following steps are necessary to have a proper environment setup, where all
...
+    ```bash
+    minikube ssh -- "sudo sed -i 's/#GatewayPorts no/GatewayPorts yes/' /etc/ssh/sshd_config && sudo systemctl reload sshd"
+    ```
+
+1. Enable tunneling from the minikube node to the cf-operator on localhost (required for webhooks)
+    ```bash
+    export ssh_server_user=docker
+    ```
+
+1. Ensure `GO111MODULE` is set
```
manno: Is that still needed? I use go 1.13 and don't set this variable. The `vendor` folder is still very helpful for docker builds, even with `GOPROXY` set.

docs/testing.md
```diff
@@ -139,37 +139,53 @@ The following steps are necessary to have a proper environment setup, where all
...
+1. Export the `OPERATOR_TEST_STORAGE_CLASS` env variable
 
     ```bash
     export OPERATOR_TEST_STORAGE_CLASS=standard
     ```
 
     _**Note**_: Require for the PVC test creation, in minikube.
 
-5. Ensure `GO111MODULE` is set
+1. Copy the ssh-key required for tunneling to /tmp
```
manno: This is because `minikube ip` is used above? I don't set up any tunneling on Linux, but  export my public interface ip instead:

```
export CF_OPERATOR_WEBHOOK_SERVICE_HOST=$(ip -4 a s dev `ip r l 0/0 | cut -f5 -d' '` | grep -oP 'inet \K\S+(?=/)')
```

jkbschmid: I went for the tunneling, because I saw that it is already part of the integration test setup (https://github.com/cloudfoundry-incubator/cf-operator/blob/d053b0eca5dfb07f83657fdd9ee85c74a269c7d5/integration/environment/operator.go#L61).
If `ssh_server_user` is set, the test framework expects the ssh-key of the node to be at `/tmp/cf-operator-tunnel-identity`.

docs/testing.md
jkbschmid: I have `GO111MODULE=on` by default, so I didn't notice whether it is still required or not. I can remove it, if you prefer.

jkbschmid:
> I also started docs for KinD: [c1e8408](https://github.com/cloudfoundry-incubator/cf-operator/commit/c1e8408ae0129e9b2a961054fd95bdcba0123535)

Good to know. Which setup do you use, personally? KinD or minikube?

docs/testing.md
manno: It defaults to `auto` and detect the `go.mod` and just work.

Anyhow, it's not really something you changed in the docs, so we can do this later.

docs/testing.md
manno: Ah, we only use the tunnel because we have to, there is no direct connection possible between the CI container and the kube cluster. I don't use it when testing locally.

However that's very interesting, can you put it in a separate section, or mention it's optional?

manno:
> > I also started docs for KinD: [c1e8408](https://github.com/cloudfoundry-incubator/cf-operator/commit/c1e8408ae0129e9b2a961054fd95bdcba0123535)
> 
> Good to know. Which setup do you use, personally? KinD or minikube?

I switched to KinD a few weeks ago. It's a little bit faster to set up and no more VM. 
`kind load image` is annoying, but not as invasive as switching my docker env to  minikubes docker.

docs/testing.md
jkbschmid: I just used the tunnel, because exporting the IP you get with `minikube ssh -- "cat /etc/resolv.conf | grep nameserver | awk '{ printf \$2 }'` didn't work for me. I just tested it with the correct IP and it's way simpler. I will adjust the doc.

PR #715, opened 2019-11-19
Create PVC in the bpm reconciler again

aduffeck:
This reverts part of 573238c and f33124f. Without this code there are
dangling PVC references in the qsts pods.

### DIFF
--- a/pkg/kube/controllers/boshdeployment/bpm_reconciler.go
+++ b/pkg/kube/controllers/boshdeployment/bpm_reconciler.go
@@ -275,6 +275,16 @@ func (r *ReconcileBPM) deployInstanceGroups(ctx context.Context, instance *bdv1.
 		log.Debugf(ctx, "Service '%s' has been %s", svc.Name, op)
 	}
 
+	// Create a persistent volume claims for containers of the instance group
+	// Right now, only one pvc is being created at /var/vcap/store
+	for _, persistentVolumeClaim := range resources.PersistentVolumeClaims {
+		err := r.createPersistentVolumeClaim(ctx, &persistentVolumeClaim)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


phil9909:
Are you sure we need this?
I changed it, because it did not work when scaling. With my change the creation of `PVC`s is delegated to the `ExtendedStatefulSet`-Controller, which will create a `PVC` for every replica and the `PodMutator` updates the `Pod` spec to reference the correct `PVC`.

aduffeck:
When trying to deploy kubecf on kind locally I ran into the problem that the PVCs were not being created. Bringing back this code fixed the problem for me.
Maybe there's something missing to make the new approach work in all cases? Can you point me to the code that's creating the PVCs for you?

phil9909:
Relevant code:

https://github.com/cloudfoundry-incubator/cf-operator/blob/bc3bd206d885765f9f09aba53b2d510edb8f57a9/pkg/kube/controllers/quarksstatefulset/quarksstatefulset_reconciler.go#L122-L128

Relevant documentation: 
https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/controllers/quarks_statefulset.md#volume-management

I have never tested it with a `kind` cluster.
Do you see any Pods or StatefulSets containing `volume-management` in their name?

f0rmiga:
@aduffeck Is it possible that you hit a bug that I fixed here? https://github.com/SUSE/kubecf/pull/166
It was introduced a few hours earlier yesterday here https://github.com/SUSE/kubecf/pull/153.

aduffeck:
@f0rmiga no, SUSE/kubecf#153 is not part of my branch. After talking to @phil9909 it looks like the pod mutating webhook that's supposed to configure the proper pvc isn't working as expected in my case. I'll continue looking into it tomorrow.

aduffeck:
For whatever reason I can not reproduce the issue that made this change necessary anymore with a fresh kind cluster. I'll retract the pr and assume it was a local issue for now.

PR #716, opened 2019-11-20
Implement explicit "quarks links" from NK -> B

edwardstudy:
One part of quarks link:
[#169419605](https://www.pivotaltracker.com/story/show/169419605)

Introduced one annotation for link provider name, and reused deployment-name label to list secrets
The cf-o collects all providers from instance group's `consumes` and does the following:
- If this provider can be found from other instance groups, remove this item
- If this provider **cannot** be found, list secrets under the same namespace
  - If the secret annotation specifies one provider, mount this secret to IG/BPM QJobs
  - If no secret or multiple secrets are matched, return an error

Until all links secrets found, the controller creates relative IG/BPM QJobs.

And we enable `UpdateOnChange` for mounted secrets, cf-o also supports restarting IG/BPM QJobs when a "link secret" changes.

### DIFF
--- a/integration/bosh_links_test.go
+++ b/integration/bosh_links_test.go
@@ -3,7 +3,6 @@ package integration_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
 	corev1 "k8s.io/api/core/v1"
 
 	bm "code.cloudfoundry.org/cf-operator/testing/boshmanifest"
@@ -65,6 +64,39 @@ var _ = Describe("BOSHLinks", func() {
 				secret, err := env.GetSecret(env.Namespace, "link-test-nats")
 				Expect(err).NotTo(HaveOccurred())
 				Expect(secret.Data).Should(HaveKeyWithValue("nats.nuts", []byte("{\"nats\":{\"password\":\"changeme\",\"port\":4222,\"user\":\"admin\"}}")))
+			})
+		})
+	})
+
+	Context("when deployment has explicit external link dependencies", func() {
+		BeforeEach(func() {
+			natsConfigMap := env.Catalog.NatsConfigMap(deploymentName)
+			tearDown, err := env.CreateConfigMap(env.Namespace, natsConfigMap)
+			Expect(err).NotTo(HaveOccurred())
+			tearDowns = append(tearDowns, tearDown)
+
+			natsSecret := env.Catalog.NatsSecret(deploymentName)
+			tearDown, err = env.CreateSecret(env.Namespace, natsSecret)
+			Expect(err).NotTo(HaveOccurred())
+			tearDowns = append(tearDowns, tearDown)
+
+			natsPod := env.Catalog.NatsPod(deploymentName)
+			tearDown, err = env.CreatePod(env.Namespace, natsPod)
+			Expect(err).NotTo(HaveOccurred())
+			tearDowns = append(tearDowns, tearDown)
+
+			natsService := env.Catalog.NatsService(deploymentName)
+			tearDown, err = env.CreateService(env.Namespace, natsService)
+			Expect(err).NotTo(HaveOccurred())
+			tearDowns = append(tearDowns, tearDown)
+
+			boshManifest = env.BOSHManifestSecret(manifestRef, bm.NatsSmokeTestWithExternalLinks)
+		})
+
+		It("creates a secret for each link", func() {
+			By("waiting for job rendering done", func() {
+				err := env.WaitForPods(env.Namespace,"quarks.cloudfoundry.org/instance-group-name=nats-smoke-tests")
+				Expect(err).NotTo(HaveOccurred())
 			})
 		})
 	})
--- a/pkg/bosh/converter/job_factory.go
+++ b/pkg/bosh/converter/job_factory.go
@@ -6,7 +6,6 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
-
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1b1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -159,7 +158,7 @@ func (f *JobFactory) VariableInterpolationJob(manifest bdm.Manifest) (*qjv1a1.Qu
 }
 
 // InstanceGroupManifestJob generates the job to create an instance group manifest
-func (f *JobFactory) InstanceGroupManifestJob(manifest bdm.Manifest, initialRollout bool) (*qjv1a1.QuarksJob, error) {
+func (f *JobFactory) InstanceGroupManifestJob(manifest bdm.Manifest, linkSecrets map[string]string, initialRollout bool) (*qjv1a1.QuarksJob, error) {
 	containers := []corev1.Container{}
 	ct := containerTemplate{
 		manifestName:   desiredManifestName(manifest.Name),
@@ -170,6 +169,14 @@ func (f *JobFactory) InstanceGroupManifestJob(manifest bdm.Manifest, initialRoll
 
 	linkOutputs := map[string]string{}
 
+	// Volume mounts for explicit linking provider links
+	linkVolumes := []corev1.Volume{}
+	linkVolumeMounts := []corev1.VolumeMount{}
+	for secretName, providerName := range linkSecrets {
+		linkVolumes = append(linkVolumes, linkVolume(secretName))
+		linkVolumeMounts = append(linkVolumeMounts, linkVolumeMount(secretName, providerName))
+	}
+
 	for _, ig := range manifest.InstanceGroups {
 		if ig.Instances != 0 {
 			// Additional secret for BOSH links per instance group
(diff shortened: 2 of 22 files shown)


pkg/bosh/converter/job_factory.go
```diff
@@ -152,51 +152,67 @@ func (f *JobFactory) VariableInterpolationJob(manifest bdm.Manifest) (*qjv1a1.Qu
...
 
 // InstanceGroupManifestJob generates the job to create an instance group manifest
-func (f *JobFactory) InstanceGroupManifestJob(manifest bdm.Manifest) (*qjv1a1.QuarksJob, error) {
+func (f *JobFactory) InstanceGroupManifestJob(manifest bdm.Manifest, linkSecrets map[string]string) (*qjv1a1.QuarksJob, error) {
 	containers := []corev1.Container{}
 
 	// QuarksJob will always pick the latest version for versioned secrets
 	desiredManifestName := names.DesiredManifestName(manifest.Name, "1")
 
+	// Volume mounts for explicit linking provider links
```
manno:
```suggestion
	// volumes for quarks links from native kube components
```

I tried to use words from the story.

pkg/bosh/converter/job_factory.go
```diff
@@ -152,51 +152,67 @@ func (f *JobFactory) VariableInterpolationJob(manifest bdm.Manifest) (*qjv1a1.Qu
...
+	return f.gatheringJob(qJobName, manifest, desiredManifestName, names.DeploymentSecretTypeInstanceGroupResolvedProperties, containers, linkVolumes)
 }
 
 // BPMConfigsJob returns an quarks job to calculate BPM information
-func (f *JobFactory) BPMConfigsJob(manifest bdm.Manifest) (*qjv1a1.QuarksJob, error) {
+func (f *JobFactory) BPMConfigsJob(manifest bdm.Manifest, linkSecrets map[string]string) (*qjv1a1.QuarksJob, error) {
 	containers := []corev1.Container{}
 	desiredManifestName := names.DesiredManifestName(manifest.Name, "1")
 
+	// Volume mounts for explicit linking provider links
```
manno: Comment like above.
This looks like duplication.

Maybe this can be calculated somewhere in the caller? Would it be useful to pass a type instead of `map[string]string`?

pkg/bosh/converter/volume_factory.go
```diff
@@ -496,3 +499,22 @@ func generateVolumeName(secretName string) string {
 	}
 	return names.Sanitize(volName)
 }
+
+func linkVolume(secretName string) corev1.Volume {
```
manno: I find the name to generic. If these methods were on a struct, this could read `quarksLinks.volume()` or something.

pkg/bosh/manifest/cmd_instance_group_resolver.go
```diff
@@ -113,6 +113,58 @@ func (dg *InstanceGroupResolver) Manifest() (Manifest, error) {
...
+					case "type":
+						linkType = string(varBytes)
+					default:
+						properties[propertyFileName] = string(varBytes)
+					}
+				}
+				return nil
+			})
+
+			// TODO How to find linkAddress/jobsInstances
```
manno: Don't commit TODOs

pkg/bosh/manifest/cmd_instance_group_resolver.go
```diff
@@ -113,6 +113,58 @@ func (dg *InstanceGroupResolver) Manifest() (Manifest, error) {
...
+			dg.jobProviderLinks[linkType][linkName] = JobLink{
+				//Address:    linkAddress,
+				//Instances:  jobsInstances,
```
manno: Commented code

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
```diff
@@ -28,8 +31,8 @@ import (
 // JobFactory creates Jobs for a given manifest
 type JobFactory interface {
 	VariableInterpolationJob(manifest bdm.Manifest) (*qjv1a1.QuarksJob, error)
-	InstanceGroupManifestJob(manifest bdm.Manifest) (*qjv1a1.QuarksJob, error)
-	BPMConfigsJob(manifest bdm.Manifest) (*qjv1a1.QuarksJob, error)
+	InstanceGroupManifestJob(manifest bdm.Manifest, linkSecrets map[string]string) (*qjv1a1.QuarksJob, error)
+	BPMConfigsJob(manifest bdm.Manifest, providerSecrets map[string]string) (*qjv1a1.QuarksJob, error)
```
manno: Provider secrets reads ambiguous to me. Could be a cloud provider.

pkg/bosh/manifest/cmd_instance_group_resolver.go
```diff
@@ -113,6 +113,58 @@ func (dg *InstanceGroupResolver) Manifest() (Manifest, error) {
 	return igManifest, nil
 }
 
+// CollectQuarksLinks collect all links from path
+func (dg *InstanceGroupResolver) CollectQuarksLinks(linksPath string) error {
+	links, err := ioutil.ReadDir(linksPath)
```
manno: This should use `afero.Fs` at some point, so we can test it

pkg/bosh/converter/job_factory.go
edwardstudy: About second points, I think it's better that `linkVolume` and `linkVolumeMount` stay in converter package.

pkg/bosh/manifest/cmd_instance_group_resolver.go
edwardstudy: I changed my code. But we still had some io ops from ioutil in this package.

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
edwardstudy: Renamed

edwardstudy:
Following https://github.com/cloudfoundry-incubator/cf-operator/blob/cc30ffa6d9793472a2e958db1b3bab403c5d0794/docs/entanglements.md
- collect services/pods infos in controller and collect other link infos in IG jobs.

pkg/bosh/manifest/cmd_instance_group_resolver.go
edwardstudy: Done

integration/bosh_links_test.go
```diff
@@ -3,7 +3,6 @@ package integration_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
```
manno: nitpicking: we're ordering imports in groups: core,github,k8s,our

integration/bosh_links_test.go
```diff
@@ -68,4 +67,37 @@ var _ = Describe("BOSHLinks", func() {
...
+
+			natsService := env.Catalog.NatsService(deploymentName)
+			tearDown, err = env.CreateService(env.Namespace, natsService)
+			Expect(err).NotTo(HaveOccurred())
+			tearDowns = append(tearDowns, tearDown)
+
+			boshManifest = env.BOSHManifestSecret(manifestRef, bm.NatsSmokeTestWithExternalLinks)
+		})
+
+		It("creates a secret for each link", func() {
```
manno: The test checks for pods however? Should it check for secrets? Or should this read `"waits for pods from the instance gr out"`?

pkg/bosh/converter/job_factory.go
```diff
@@ -6,7 +6,6 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
-
```
manno: nitpicking again

pkg/bosh/manifest/manifest.go
```diff
@@ -134,19 +134,19 @@ type AddOn struct {
...
-	DNS            DomainNameService        `json:"-"`
+	Name           string                 `json:"name"`
+	DirectorUUID   string                 `json:"director_uuid"`
+	InstanceGroups InstanceGroups         `json:"instance_groups,omitempty"`
+	Features       *Feature               `json:"features,omitempty"`
+	Tags           map[string]string      `json:"tags,omitempty"`
+	Releases       []*Release             `json:"releases,omitempty"`
+	Stemcells      []*Stemcell            `json:"stemcells,omitempty"`
+	AddOns         []*AddOn               `json:"addons,omitempty"`
+	Properties     map[string]interface{} `json:"properties,omitempty"`
```
manno: wow, this was an array before, how did that work..

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
```diff
@@ -217,3 +239,203 @@ func (r *ReconcileBOSHDeployment) createQuarksJob(ctx context.Context, instance
 
 	return err
 }
+
+// listLinkInfos returns a LinkInfos containing link providers if needed
+// and updates `quarks_links` properties
+func (r *ReconcileBOSHDeployment) listLinkInfos(instance *bdv1.BOSHDeployment, manifest *bdm.Manifest) (converter.LinkInfos, error) {
```
manno: I'll say more about this in the next retro, but I think we're all putting too much code in the 'reconciler.go' files.
The funcs which don't need a context, could often live with the structs they modify. 
The funcs that use a context could be separate interfaces, like the job_creator.go in quarks-job.

integration/bosh_links_test.go
```diff
@@ -68,4 +67,37 @@ var _ = Describe("BOSHLinks", func() {
 			})
 		})
 	})
+
+	Context("when deployment has explicit external link dependencies", func() {
+		BeforeEach(func() {
+			natsConfigMap := env.Catalog.NatsConfigMap(deploymentName)
```
manno: just `env.NatsConfigMap` is enough, because of the embedded struct

pkg/kube/controllers/boshdeployment/deployment_reconciler.go
edwardstudy: A good point, thx.

PR #719, opened 2019-11-21
Fix 'set -e' and others get lost when sourcing 'bin/include/versioning'

phil9909:
Currently when I run
```bash
make build-image
echo $?
```
it will echo `0` even if the docker build failes.

### Problem
inside `build-image` the `errexit` option is set correctly.
https://github.com/cloudfoundry-incubator/cf-operator/blob/bc3bd206d885765f9f09aba53b2d510edb8f57a9/bin/build-image#L3

But when sourcing `bin/include/versioning` the option gets resetted:
https://github.com/cloudfoundry-incubator/cf-operator/blob/bc3bd206d885765f9f09aba53b2d510edb8f57a9/bin/include/versioning#L20

### Proposed solution
In the beginning of `bin/include/versioning` current shell-options are stored and restored in the end.

This PR also changes how the `XTRACE` env varaible is evaluated.

### DIFF
--- a/bin/include/versioning
+++ b/bin/include/versioning
@@ -1,10 +1,11 @@
 #!/bin/bash
 
-set +o errexit +o nounset
-
-test -n "${XTRACE}" && set -o xtrace
-
+old_options="$(set +o); set -$-"
 set -o errexit -o nounset
+
+if [ -n "${XTRACE:-}" ] ; then
+    set -o xtrace
+fi
 
 GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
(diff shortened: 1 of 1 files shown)


manno:
Half of that include file is now dealing with error handling. I'm wondering if setting shell options is better left to the consuming script?
I briefly tried it, maybe we can remove all the `set` commands from the include files?

phil9909:
Yes, we could also do this.
Should I update the PR or will you do it (seems like you're half way done)?

manno:
> Yes, we could also do this.
> Should I update the PR or will you do it (seems like you're half way done)?

I tried it and other scripts break :) I'll try to add to this PR or start a new one ..

viovanov:
closed in favor of #720

Issue #730, opened 2019-12-02
cf-operator crashes on first run


f0rmiga:
It fails with:
```
2019/12/02 20:01:00 cf-operator command failed. Failed to create new manager.: failed to add controllers to manager: no matches for kind "ExtendedStatefulSet" in version "quarks.cloudfoundry.org/v1alpha1"
```

rohitsakala:
Hi @f0rmiga ,

I tried to reproduce this on latest master in `cf-operator`. Steps I tried

1. Deleted my kind clusters
2. Created a new kind cluster.
3. Ran `go run ./cmd/cf-operator/main.go
4. and there was no error as above.

Am I missing anything ?

f0rmiga:
@rohitsakala Try deploying the helm chart.

rohitsakala:
Ran deploying using helm chart for 10 times in a loop. Couldn't reproduce. @f0rmiga to provide more info if seen again.

manno:
I'm going to close the issue. If this happens again, @f0rmiga please re-open and add more information on how to reproduce it: 
Which commands did you run, what's kind of Kubernetes cluster are you using, what version of the helm chart was used, etc.

PR #732, opened 2019-12-05
Rename ExtendedStatefulSet to QuarksStatefulSet

rohitsakala:


rohitsakala:
> Did this fix any issue?

No. actually. I was trying to fix this issue [#730](https://github.com/cloudfoundry-incubator/cf-operator/issues/730) and observed these are still not renamed. So put a PR.

PR #738, opened 2019-12-08
Introduce active-passive controller

qu1queee:
[#169117521](https://www.pivotaltracker.com/story/show/169117521)
This controller reconciles on events from Quarksstatefulsets instances

The main logic around this controller is that it will execute a probe
based on a container name per pod. When the probe cmd executions succeeds,
it will label the pod with a `pod-designation` key, with a `true` value.

For pods, where the probe fails, if the label exists, it will be removed.

This PR contains the following:

- Adds a new controller to the Manager
- Adds a new active-passive controller under the `kube/controllers/quarkstatefulset` dir
- Adds a new active-passive reconciler for the above controller.
- Updates quarksstatefulset types
  - Introduces a new field call `ActivePassiveProbe`, which holds a value of the type
    `map[string]*corev1.Probe`, which basically reuses the existing k8s Probe functionality
  - The Probe `PeriodSeconds` field, will determine how often the active-passive controller
    will be reconciling.
- Update quarksstatefulset generated fake files
  - Due to new field for the CRD struct
- Adds only integration tests for this controller
  - New set of supportive funcs for the Machine interface
    This includes:
      - logic for events lookup
      - pod cmd execution in containers
      - ability to patch pods
  - added a new `quarks_statefulset_active_passive_test.go`
    - holds different scenarios to validate the crd behaviour.
- Adds a new sample document, with a ActivePassive Probe

### DIFF
--- a/docs/examples/quarks-statefulset/qstatefulset_active_passive.yaml
+++ b/docs/examples/quarks-statefulset/qstatefulset_active_passive.yaml
@@ -1,0 +1,31 @@
+apiVersion: quarks.cloudfoundry.org/v1alpha1
+kind: QuarksStatefulSet
+metadata:
+  name: example-quarks-statefulset
+spec:
+  updateOnConfigChange: true
+  activePassiveProbe:
+    busybox:
+      periodSeconds: 5
+      exec:
+        command:
+        - /bin/sh
+        - -c
+        - date
+  template:
+    metadata:
+      labels:
+        app: example-statefulset
+    spec:
+      replicas: 1
+      template:
+        metadata:
+          labels:
+            app: example-statefulset
+        spec:
+          containers:
+          - name: busybox
+            image: busybox
+            command:
+            - sleep
+            - "3600"
--- a/e2e/kube/examples_test.go
+++ b/e2e/kube/examples_test.go
@@ -113,6 +113,23 @@ var _ = Describe("Examples Directory", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("it labels the first pod as active", func() {
+			yamlUpdatedFilePath := examplesDir + "quarks-statefulset/qstatefulset_active_passive.yaml"
+			By("Applying a quarkstatefulset with active-passive probe")
+			err := cmdHelper.Apply(namespace, yamlUpdatedFilePath)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = wait.PollImmediate(time.Second*5, time.Second*35, func() (bool, error) {
+				err := kubectl.WaitForPod(namespace, "quarks.cloudfoundry.org/pod-active", "example-quarks-statefulset-0")
+				if err != nil {
+					return false, err
+				}
+				return true, nil
+			})
+			Expect(err).ToNot(HaveOccurred())
+
+		})
+
 	})
 
 	Context("bosh-deployment service example", func() {
--- a/integration/environment/machine_esv1.go
+++ b/integration/environment/machine_esv1.go
@@ -1,15 +1,24 @@
 package environment
 
 import (
-	"github.com/pkg/errors"
-	"k8s.io/api/apps/v1beta1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/util/wait"
+	"encoding/json"
+	"os"
 
 	qstsv1a1 "code.cloudfoundry.org/cf-operator/pkg/kube/apis/quarksstatefulset/v1alpha1"
 	"code.cloudfoundry.org/quarks-utils/testing/machine"
+	"github.com/pkg/errors"
+	"k8s.io/api/apps/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	clientscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
 )
 
 // GetStatefulSet gets a StatefulSet custom resource
@@ -78,6 +87,46 @@ func (m *Machine) UpdateQuarksStatefulSet(namespace string, ess qstsv1a1.QuarksS
 func (m *Machine) DeleteQuarksStatefulSet(namespace string, name string) error {
 	client := m.VersionedClientset.QuarksstatefulsetV1alpha1().QuarksStatefulSets(namespace)
 	return client.Delete(name, &metav1.DeleteOptions{})
+}
+
+// WaitForPodLabelToExist blocks until the specified label appears
+func (m *Machine) WaitForPodLabelToExist(n string, podName string, label string) error {
+	return wait.PollImmediate(m.PollInterval, m.PollTimeout, func() (bool, error) {
+		return m.PodLabelToExist(n, podName, label)
+	})
+}
+
+// PodLabelToExist returns true if the label exist in the specified pod
+func (m *Machine) PodLabelToExist(n string, podName string, label string) (bool, error) {
+	pod, err := m.Clientset.CoreV1().Pods(n).Get(podName, metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+	labels := pod.GetLabels()
+	if _, found := labels[label]; found {
+		return true, nil
+	}
+	return false, nil
+}
+
+// WaitForPodLabelToNotExist blocks until the specified label is not present
+func (m *Machine) WaitForPodLabelToNotExist(n string, podName string, label string) error {
+	return wait.PollImmediate(m.PollInterval, m.PollTimeout, func() (bool, error) {
+		return m.PodLabelToNotExist(n, podName, label)
+	})
+}
+
+// PodLabelToNotExist returns true if the label does not exist
+func (m *Machine) PodLabelToNotExist(n string, podName string, label string) (bool, error) {
+	pod, err := m.Clientset.CoreV1().Pods(n).Get(podName, metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+	labels := pod.GetLabels()
+	if _, found := labels[label]; !found {
+		return true, nil
+	}
+	return false, nil
 }
 
 // WaitForStatefulSetDelete blocks until the specified statefulSet is deleted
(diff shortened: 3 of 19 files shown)


docs/examples/quarks-statefulset/qstatefulset_active_passive.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: quarks.cloudfoundry.org/v1alpha1
+kind: QuarksStatefulSet
+metadata:
+  name: example-quarks-statefulset
+spec:
+  updateOnConfigChange: true
+  activepassiveprobe:
```
viovanov: `activePassiveProbe`

e2e/kube/examples_test.go
```diff
@@ -113,6 +113,23 @@ var _ = Describe("Examples Directory", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("label the first pod as active", func() {
```
viovanov: it labels

integration/environment/machine_esv1.go
```diff
@@ -227,3 +249,80 @@ func (m *Machine) StatefulSetNewGeneration(namespace string, name string, versio
...
+		return false, errors.Wrapf(err, "failed executing command in pod: %s, container: %s in namespace: %s",
+			pod.Name,
+			container,
+			pod.Namespace,
+		)
+	}
+	return true, nil
+}
+
+// PatchPod applies a patch into an specific pod
```
viovanov: into a specific pod

pkg/kube/apis/quarksstatefulset/v1alpha1/register.go
```diff
@@ -46,6 +46,10 @@ var (
 							Type:        "boolean",
 							Description: "Indicate whether to update Pods in the StatefulSet when an env value or mount changes",
 						},
+						"activepassiveprobe": {
```
viovanov: `activePassiveProbe`

pkg/kube/apis/quarksstatefulset/v1alpha1/types.go
```diff
@@ -29,6 +30,8 @@ var (
 	LabelPodOrdinal = fmt.Sprintf("%s/pod-ordinal", apis.GroupName)
 	// LabelQStsName is the name of the QuarksStatefulSet owns this resource
 	LabelQStsName = fmt.Sprintf("%s/quarks-statefulset-name", apis.GroupName)
+	// LabelActiveContainer is the active container on an active/passive setup
+	LabelActiveContainer = fmt.Sprintf("%s/pod-designation", apis.GroupName)
```
viovanov: Designation is an ambiguous term - it can also mean "the name of something" 
I think `pod-active` makes more sense.

pkg/kube/controllers/quarksstatefulset/active_passive_controller.go
```diff
@@ -0,0 +1,82 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// AddStatefulSetActivePassive creates a new QuarksStatefulSet controller that watches multiple instances
+// of an specific resource, and decides which is active and which is passive
```
viovanov: of a specific resource

pkg/kube/controllers/quarksstatefulset/active_passive_controller.go
```diff
@@ -0,0 +1,82 @@
...
+
+	r := NewActivePassiveReconciler(ctx, config, mgr, kclient)
+
+	// Create new controller
+	c, err := controller.New("active-passive-controller", mgr, controller.Options{
+		Reconciler:              r,
+		MaxConcurrentReconciles: config.MaxQuarksStatefulSetWorkers,
+	})
+	if err != nil {
+		return errors.Wrap(err, "Adding active-passive controller to manager failed.")
```
viovanov: lower case

pkg/kube/controllers/quarksstatefulset/active_passive_controller.go
```diff
@@ -0,0 +1,82 @@
...
+				return true
+			}
+			return false
+		},
+	}
+	err = c.Watch(&source.Kind{Type: &qstsv1a1.QuarksStatefulSet{}},
+		&handler.EnqueueRequestForObject{},
+		stsPredicates)
+	if err != nil {
+		return errors.Wrapf(err, "Watching QuarksStatefulSet failed in active/passive controller.")
```
viovanov: lower case

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+	"k8s.io/client-go/rest"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/remotecommand"
+	crc "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// NewActivePassiveReconciler returns a new reconcile.Reconciler for active/passive controller
```
viovanov: for the active/passive

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+		ctx:        ctx,
+		config:     config,
+		client:     mgr.GetClient(),
+		kclient:    kclient,
+		scheme:     mgr.GetScheme(),
+		kubeconfig: mgr.GetConfig(),
+	}
+}
+
+// ReconcileStatefulSetActivePassive reconciles an QuarksStatefulSet object when references changes
```
viovanov: when references change

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+type ReconcileStatefulSetActivePassive struct {
+	ctx        context.Context
+	client     crc.Client
+	kclient    kubernetes.Interface
+	scheme     *runtime.Scheme
+	config     *config.Config
+	kubeconfig *restclient.Config
+}
+
+// Reconcile reads that state of the cluster for a QuarksStatefulSet object
```
viovanov: the state

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+// and makes changes based on the state read and what is in the QuarksStatefulSet.Spec
+// Note:
+// The Reconcile Loop will always requeue the request under completition. For this specific
```
viovanov: stop before under completion

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// Set the ctx to be Background, as the top-level context for incoming requests.
+	ctx, cancel := context.WithTimeout(r.ctx, r.config.CtxTimeOut)
+	defer cancel()
+
+	ctxlog.Info(ctx, "Reconciling for active/passive QuarksStatefulSet", request.NamespacedName)
+
+	if err := r.client.Get(ctx, request.NamespacedName, qSts); err != nil {
+		if apierrors.IsNotFound(err) {
+			// Reconcile successful - don´t requeue
```
viovanov: ` vs '

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+
+	err = r.defineActiveContainer(ctx, containerName, ownedPods, qSts)
+	if err != nil {
+		// Reconcile failed due to error - requeue
+		return reconcile.Result{}, err
+	}
+
+	periodSeconds := time.Second * time.Duration(qSts.Spec.ActivePassiveProbe[containerName].PeriodSeconds)
+	if periodSeconds == (time.Second * time.Duration(0)) {
+		ctxlog.WithEvent(qSts, "active-passive").Debugf(ctx, "periodSeconds probe was not specified, going to default to 10 secs")
```
viovanov: message says 10, result says 5
i think it should be 30

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+		periodSeconds = time.Second * 5
+	}
+
+	// Reconcile for any reason than error after the ActivePassiveProbe PeriodSeconds
+	return reconcile.Result{RequeueAfter: periodSeconds}, nil
+}
+
+func (r *ReconcileStatefulSetActivePassive) defineActiveContainer(ctx context.Context, container string, pods *corev1.PodList, qSts *qstsv1a1.QuarksStatefulSet) (err error) {
```
viovanov: rename to "markActiveContainers"

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+	var p *corev1.Pod
+	labelledPods := r.getActiveLabels(pods, container)
+
+	probeCmd := qSts.Spec.ActivePassiveProbe[container].Exec.Command
+
+	// switch depending on pods with active label
+	switch pl := len(labelledPods); {
+	// Nothing have been labeled yet
```
viovanov: nothing has been

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+	case pl == 0:
```
viovanov: we should always check all pods - more than one pod could be active

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+		podList,
+		crc.InNamespace(desiredSts.Namespace),
+	)
+	if err != nil {
+		return nil, err
+	}
+	return podList, nil
+}
+
+func (r *ReconcileStatefulSetActivePassive) listStatefulSetVersions(ctx context.Context, qStatefulSet *qstsv1a1.QuarksStatefulSet) (map[int]bool, error) {
```
viovanov: `getStatefulSet`

should be a helper of qsts

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+		SubResource("exec").
+		VersionedParams(&corev1.PodExecOptions{
+			Container: container,
+			Command:   command,
+			Stdin:     true,
+			Stdout:    true,
+			Stderr:    false,
+		}, clientscheme.ParameterCodec)
+
+	executor, err := remotecommand.NewSPDYExecutor(r.kubeconfig, "POST", req.URL())
```
viovanov: is this the only way? the kubeconfig part looks strange

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+		if !found {
+			return versions, errors.Errorf("version annotation is not found from: %+v", statefulSet.Annotations)
+		}
+
+		version, err := strconv.Atoi(strVersion)
+		if err != nil {
+			return versions, errors.Wrapf(err, "version annotation is not an int: %s", strVersion)
+		}
+
+		ready, err := r.isStatefulSetReady(ctx, &statefulSet)
```
viovanov: shouldn't be required - there's 1 sts per qsts now - should just be able to retrieve it

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+	maxAvailableVersion := qSts.GetMaxAvailableVersion(statefulSetVersions)
+
+	statefulSets, err := listStatefulSetsFromInformer(ctx, r.client, qSts)
+	if err != nil {
+		// Reconcile failed due to error - requeue
+		return reconcile.Result{}, errors.Wrapf(err, "couldn't list StatefulSets for active/passive reconciliation.")
+	}
+
+	// Retrieve the last versioned statefulset, based on the max available version
+	desiredSts, err := getDesiredSts(statefulSets, fmt.Sprintf("%d", maxAvailableVersion))
```
viovanov: shouldn't be required anymore

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
qu1queee: Yes, I added support for this

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
qu1queee: we do not longer need this function, I removed it

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
qu1queee: maybe you got confuse, its actually a rest.Config

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
qu1queee: yes, that simplify my code

qu1queee:
@viovanov I addressed all your comments, asking to re-review my PR

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
viovanov: You're still only making on pod active at once.
I see you're always making sure there's only one active pod at the same time.

I don't think we need to do this - we should just run the probe and mark accordingly (for all pods, all the time).

pkg/kube/controllers/quarksstatefulset/active_passive_reconciler.go
```diff
@@ -0,0 +1,247 @@
...
+
+	probeCmd := qSts.Spec.ActivePassiveProbe[container].Exec.Command
+
+	for _, pod := range pods.Items {
+		ctxlog.WithEvent(qSts, "active-passive").Debugf(
+			ctx,
+			"validating probe in pod: %s",
+			pod.Name,
+		)
+		succeed, _ := r.execContainerCmd(&pod, container, probeCmd)
```
viovanov: You're ignoring err here - I'm assuming err is not nil if there was an api communication failure for example.
We should log this.

PR #739, opened 2019-12-09
Use BPM memory limits for container resource limits

kramerul:
It turned out that doppler uses a lot of memory in a setup with 2 replicas. 
Therefore, we would like be able to limit the memory usage. Using the existing BPM configuration variables for container limits would help.


pkg/bosh/converter/container_factory.go
```diff
@@ -515,6 +518,15 @@ func bpmProcessContainer(
...
 	command, args := generateBPMCommand(&process, postStart)
+	limits := corev1.ResourceList{}
+	if process.Limits.Memory != "" {
+		quantity, err := resource.ParseQuantity(process.Limits.Memory)
+		if err != nil {
+			log.Errorf(context.TODO(), "Error parsing %s: %v", process.Limits.Memory, err)
+		} else {
+			limits[corev1.ResourceMemory] = quantity
+		}
+	}
```
loewenstein: Just wondering, why the `limits` don't come from where the `requests` come from.

kramerul: The `Limits` structure in the `BPMProcess` was available before our change for the `Requests`. Therefore, I decided to use the data provided by `BPMProcess.Limits`

viovanov:
@kramerul lgtm - can you please add a test?

kramerul:
@viovanov , I added a negative test. A test was already included in the existing BPM resource test.

PR #742, opened 2019-12-10
These helm defaults are not used by the chart

manno:
[#169418463](https://www.pivotaltracker.com/story/show/169418463)


manno:
> Do you know the reason why this values end-up here?

They are from the initial commit, which added a helm chart. `helm create` generates those.
They are just example values and we deleted all templates that used them.

PR #757, opened 2019-12-19
Refactor timeouts in statefulset-rollout-controller

jkbschmid:
Extract method to handle timeouts.
Fix small/unlikely bug to differentiate between 0 seconds left and do not timeout at all.


pkg/kube/controllers/statefulset/statefulset_rollout_reconciler.go
```diff
@@ -164,17 +158,28 @@ func (r *ReconcileStatefulSetRollout) Reconcile(request reconcile.Request) (reco
...
 			return reconcile.Result{}, err
 		}
 	}
 	return resultWithRetrigger, nil
 }
 
-func (r *ReconcileStatefulSetRollout) updateStatefulSet(ctx context.Context, statefulset appsv1.StatefulSet, oldPartition int32) error {
-	meltdown.SetLastReconcile(&statefulset.ObjectMeta, time.Now())
-	err := r.update(ctx, &statefulset)
+func (r *ReconcileStatefulSetRollout) failIfTimedOut(ctx context.Context, statefulSet appsv1.StatefulSet, timeout string) (bool, error) {
```
edwardstudy: About this func signature, could we return the error only? And return an error wrap ("timeout error") if timed out at this line: https://github.com/cloudfoundry-incubator/cf-operator/pull/757/files#diff-87ec6f137a4c3ba48774134d91848bd2R176

I have no idea about this first bool value can explain what status.

jkbschmid: We have to distinguish three cases: 
- timedOut and error -> exit and retry reconciliation to set 'Failed'-Status in the StatefulSet
- timedOut and no error -> exit and do not retry reconciliation
- not timedOut and no error -> continue reconciliation
Thus, we need the timedOut so we don't retry indefinitely. 
Alternatively, we inline the method like so
```suggestion
	if getTimeOut(ctx, statefulSet, AnnotationUpdateWatchTime) < 0 {
		statefulSet.Annotations[AnnotationCanaryRollout] = rolloutStateFailed
		if err := r.updateStatefulSet(ctx, &statefulSet); err != nil {
			ctxlog.Debug(ctx, "Error updating StatefulSet ", statefulSet.Name, err)
		}
		return reconcile.Result{}, err
	}
```

manno: I'm fine with the original implementation. It's however easy to overlook, that returning an error which is not nil, will retry the reconcile.

PR #765, opened 2020-01-07
Fix flaky entangled pod and add failure messages to statefulset e2e tests

manno:
[#170573718](https://www.pivotaltracker.com/story/show/170573718)


manno:
> I guessed it's a enhancement for quarksStatefulset e2e testing not entangled pod, right?

Oh, right. It's both. The entangled pod fix is really sneaky, just bumping quarks-utils to get the new WaitForData func.

I'll add more info to the commit messages.

PR #767, opened 2020-01-08
Combine bpm and ig manifest jobs into one

manno:
Running in one container, but still creates two separate secrets

[#170257050](https://www.pivotaltracker.com/story/show/170257050)

Requires: https://github.com/cloudfoundry-incubator/quarks-job/pull/38


docs/commands/cf-operator_util_instance-group.md
```diff
@@ -7,6 +7,7 @@ Resolves instance group properties of a BOSH manifest
 Resolves instance group properties of a BOSH manifest.
 
 This will resolve the properties of an instance group and return a manifest for that instance group.
+Also calculates and prints the BPM configurations for all all BOSH jobs of that instance group.
```
viovanov: double "all"

pkg/bosh/converter/job_factory.go
```diff
@@ -302,14 +283,26 @@ func (f *JobFactory) releaseImageQJob(name string, manifest bdm.Manifest, secret
...
+			InstanceGroupOutputFilename: qjv1a1.SecretOptions{
+				// the same as names.InstanceGroupSecretName(secretType, manifestName, container.Name, "")
+				Name: igPrefix + container.Name,
+				AdditionalSecretLabels: map[string]string{
+					bdv1.LabelDeploymentSecretType: names.DeploymentSecretTypeInstanceGroupResolvedProperties.String(),
+				},
+				Versioned: true,
+			},
+			BPMOutputFilename: qjv1a1.SecretOptions{
+				Name: bpmPrefix + container.Name,
```
viovanov: Shouldn't this name be constructed in `names`?

manno: the base is constructed in `names`, inside the loop the container name is added. I just copied that behavior from the other output. I think it's an optimization, so I'd leave it at that for now.

PR #778, opened 2020-01-15
Add missing default values to the helm chart and comments

rohitsakala:
[#169119949](https://www.pivotaltracker.com/n/projects/2192232/stories/169119949)

### DIFF
--- a/deploy/helm/cf-operator/values.yaml
+++ b/deploy/helm/cf-operator/values.yaml
@@ -1,41 +1,86 @@
-# Default values for cf-operator.
+## Default values for Cf-operator Helm Chart.
+## This is a YAML-formatted file.
+## Declare variables to be passed into your templates.
+
+
+# applyCRD is a boolean to control the installation of CRD's.
+applyCRD: true
+
+cluster:
+  # domain is the the Kubernetes cluster domain
+  domain: "cluster.local"
+
+# fullnameOverride overrides the release name
+fullnameOverride: ""
+
+# image is the docker image of quarks job.
 image:
+  # repository that provides the operator docker image.
   repository: cf-operator
+  # org that provides the operator docker image.
   org: cfcontainerization
+  # tag of the operator docker image
   tag: foobar
+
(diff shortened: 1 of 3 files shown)


docs/development.md
```diff
@@ -310,6 +310,9 @@ docker tag <quarks-job-tag> cfcontainerization/quarks-job:$QUARKS_JOB_IMAGE_TAG
 
 * In PR reviews, make sure only reviewed branches of dependencies are used.
 * If several PRs across repos belong together, be careful to merge them together and in order, so that the master branches stay compatible if one of the merges fails.
+* Make sure that the cf-operator repo PR has updated value of quarks image tag at these places
+  * [Dependency File](https://github.com/cloudfoundry-incubator/cf-operator/blob/master/bin/include/dependencies#L3)
+  * [Helm Chart Value](https://github.com/cloudfoundry-incubator/quarks-job/blob/master/deploy/helm/quarks-job/values.yaml#L5)
```
manno: This should not be necessary.

However I think you found a bug.

Here https://github.com/cloudfoundry-incubator/cf-operator/blob/master/bin/build-helm#L19 we should be using `QUARKS_JOB_HELM_VERSION` instead of `ARTIFACT_VERSION`.
However, we don't have a helm repo, so nobody is using that value...

rohitsakala: Done.

deploy/helm/cf-operator/values.yaml
```diff
@@ -1,41 +1,86 @@
...
+
+# image is the docker image of quarks job.
 image:
+  # repository that provides the operator docker image.
   repository: cf-operator
+  # org that provides the operator docker image.
   org: cfcontainerization
-  tag: foobar
+  # tag of the operator docker image
+  tag: v1.0.0-2.g4ceba93a
```
manno: We had a story some time ago, to explicitly set this to an invalid tag. Can you change it back to foobar? :)

rohitsakala: Done.

PR #784, opened 2020-01-21
Simple Secret Rotation

manno:
Rotate listed quarks secrets
    
[#167366400](https://www.pivotaltracker.com/story/show/167366400)


docs/examples/quarks-secret/rotate.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: rotate
+  labels:
+    quarks.cloudfoundry.org/secret-rotation: "true"
+data:
+  secrets: '["generate-password"]'
```
viovanov: 👍

integration/environment/namespace.go
```diff
@@ -38,6 +39,12 @@ func (e *Environment) SetupNamespace() error {
 	return nil
 }
 
+// NamespaceDeletionInProgress returns true if the error indicates deletion will happen
```
viovanov: very cool - so we don't get all those red herrings on stdout?

pkg/bosh/manifest/manifest.go
```diff
@@ -511,17 +509,16 @@ func (m *Manifest) ApplyAddons(ctx context.Context) error {
...
 				return errors.Wrap(err, "failed to process include placement matches")
 			}
-			exclude, err := m.addOnPlacementMatch(ctx, "exclusion", ig, addon.Exclude)
+			exclude, err := m.addOnPlacementMatch("exclusion", ig, addon.Exclude)
 			if err != nil {
 				return errors.Wrap(err, "failed to process exclude placement matches")
 			}
 
 			if exclude || !include {
-				ctxlog.Debugf(ctx, "Addon '%s' doesn't match instance group '%s'", addon.Name, ig.Name)
```
viovanov: This was really useful when kubecf was starting.
Addon match rules are complicated, and we don't have any other feedback for understanding why an addon is or isn't applied.

integration/environment/namespace.go
manno: yes, had to make one small fix though :)

pkg/bosh/manifest/manifest.go
manno: Hm, maybe we need another solution here. Maybe just passing a zap logger and logging only "interesting" cases. The addons were several thousand lines of logging.

PR #791, opened 2020-01-24
Change to secret updates deployment

manno:
* implicit secret has to be used in manifest to trigger update
* change to quarks secret, via implicit secret, was ignored
* reconciler now sets generated to false when quarks secret changed
* quarks secret now triggers directly, when status is updated to generated=false

[#169709469](https://www.pivotaltracker.com/story/show/169709469)


manno:
> Does this need some updates to docs?

I will update the QuarksSecret docs in a separate PR

PR #798, opened 2020-01-30
Consume Quarks links as environment variables

HeavyWombat:
[#170725085](https://www.pivotaltracker.com/story/show/170725085)

Using Quarks links (entanglements) on the consumer side required changes
in the operator so that a link property can be used through an
environment variable and as a file with just the property content. In
theory, this should enable consumers to retrieve the property value in
their respective project with minimal adjustments.

Introduce flattened link properties in the Kubernetes secrets that
contain the link properties. The name and the data of the link secret
change from `link-deploymentname-name`
```
"nats.nats": "{\"nats\": { \"port\": 4442}, ... }"
```
to a new flatten look and name `link-deploymentname-type-name`
```
"nats.port": "4442"
"nats.user": "admin"
```
The link type and name become part of the secret name and will not be
used in the secret data as the root key anymore. Therefore, all code
sections and tests that wait or rely on the old link secret name were
updated to use the new names. This relies on new convenience functions
to create the name for both the link secret name and the type/name pair.

The `cmd_instance_group_resolver` code and tests were updated to write
the flattened properties in the `provides.json`, which is later used by
a Quarks Job to persist the data into the link secrets.

The `job_factory` code and tests were updated to use the new "fan-out"
style of the Quarks Job project that writes the content of the
aforementioned `provides.json` into seperate link secrets, each only
containing one key/value pair. With other words, there will be one
secret per each link type/name tuple.

The `pod_mutator` code and tests were changed to not mount the link
properties as one file (`link.yaml`) anymore, but that each entry in the
link properies becomes its own file in the target container. Also, each
entry in the properties will result in an environment variable starting
with the prefix `LINK_` and the respective value. For example, the nats
username will be exposed as `LINK_NATS_USER` in the containers.


HeavyWombat:
Force pushed to fix `staticcheck` and `misspell` findings.

HeavyWombat:
It might be that the tests fail, because I have to update the `quarks-job` version not just in the Go modules but somewhere else, too. I will have a look.

rohitsakala:
> It might be that the tests fail, because I have to update the `quarks-job` version not just in the Go modules but somewhere else, too. I will have a look.

Hi,

I think you missed to update the value in the dependency file.
3rd point in [0].

[0] https://github.com/cloudfoundry-incubator/cf-operator/blob/master/docs/development.md#merging-prs

HeavyWombat:
I fixed the missing `quarks-job` update in the dependency file and updated the documentation with the new details.

Please note: I added another [commit in there with another potential improvement to a flaky test case](https://github.com/cloudfoundry-incubator/cf-operator/pull/798/commits/796c42fb8e98fc1e9be1a48f86c8d14f28090aa3) that bugged me during testing.

PR #806, opened 2020-02-04
Require watch namespace

manno:
* terminate if namespace is deleted
* use operator namespace as value in watch namespaces label

[#170918922](https://www.pivotaltracker.com/story/show/170918922)


docs/commands/cf-operator.md
```diff
@@ -31,12 +31,12 @@ cf-operator [flags]
   -w, --operator-webhook-service-host string     (CF_OPERATOR_WEBHOOK_SERVICE_HOST) Hostname/IP under which the webhook server can be reached from the cluster
   -p, --operator-webhook-service-port string     (CF_OPERATOR_WEBHOOK_SERVICE_PORT) Port the webhook server listens on (default "2999")
   -x, --operator-webhook-use-service-reference   (CF_OPERATOR_WEBHOOK_USE_SERVICE_REFERENCE) If true the webhook service is targeted using a service reference instead of a URL
-      --watch-namespace string                   (WATCH_NAMESPACE) Namespace to watch for BOSH deployments
+  -a, --watch-namespace string                   (WATCH_NAMESPACE) Act on this namespace, watch for BOSH deployments and create resources (default "staging")
```
edwardstudy: Could we use `--deployment-namespace`, so we could use a short name as `-d`? :)

rohitsakala: I like it.  :heavy_plus_sign: . if it is possible. or if we can make snatch `-w` from webhook host making it `-h`. hehe.

manno: I also wanted to rename `cf-operator-namespace` to 'webhook-namespace`, because it's only needed for the webhook configuration, i.e. the namespace where the service resides which points to the webhook server.

However we talked about it in the daily and won't do any renames as part of this story.

PR #808, opened 2020-02-05
Active passive probes

jimmykarily:
This is needed to complete this story: https://github.com/SUSE/kubecf/issues/309

### DIFF
--- a/docs/examples/quarks-statefulset/qstatefulset_active_passive.yaml
+++ b/docs/examples/quarks-statefulset/qstatefulset_active_passive.yaml
@@ -4,7 +4,7 @@ metadata:
   name: example-quarks-statefulset
 spec:
   updateOnConfigChange: true
-  activePassiveProbe:
+  activePassiveProbes:
     busybox:
       periodSeconds: 5
       exec:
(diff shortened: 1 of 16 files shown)


pkg/bosh/manifest/containerization.go
```diff
@@ -9,17 +9,18 @@ import (
...
+	Release             string                   `json:"release"`
+	BPM                 *bpm.Config              `json:"bpm,omitempty" yaml:"bpm,omitempty"`
+	Ports               []Port                   `json:"ports"`
+	Run                 RunConfig                `json:"run"`
+	PreRenderScripts    PreRenderScripts         `json:"pre_render_scripts" yaml:"pre_render_scripts"`
+	PostStart           PostStart                `json:"post_start"`
+	Debug               bool                     `json:"debug" yaml:"debug"`
+	IsAddon             bool                     `json:"is_addon" yaml:"is_addon"`
+	Envs                []corev1.EnvVar          `json:"envs" yaml:"envs"`
+	ActivePassiveProbes map[string]*corev1.Probe `json:activePassiveProbes,omitempty`
```
manno: I think this would be fine without the pointer, there is no need to have an explicit nil in that list?

docs/examples/quarks-statefulset/qstatefulset_active_passive.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: example-quarks-statefulset
 spec:
   updateOnConfigChange: true
-  activePassiveProbe:
+  activePassiveProbes:
```
viovanov: API change - means we need a major release

PR #820, opened 2020-02-20
Update docs re: entanglements

mook-as:
The link is mounted as a secret where the type and name are separated by a dash, not a dot.  All of the tests are already testing this behaviour.

This is derived from `quarks-utils/pkg/names.QuarksLinkSecretKey` via `cf-operator/pkg/kube/controllers/quarkslink.link.String()`.


docs/entanglements.md
```diff
@@ -86,7 +86,7 @@ If a pod is annotated with the following:
 
 The operator will mutate the pod to:
 
-- mount the link secrets as `/quarks/link/DEPLOYMENT/<type>.<name>/<key>`
+- mount the link secrets as `/quarks/link/DEPLOYMENT/<type>-<name>/<key>`
 - add an environment variable for each key in the secret data mapping: `LINK_<key>`
 
 The `<name>` and `<type>` are the respective link type and name. For example, the nats release uses `nats` for both the name and the type of the link. The `<key>` describes the BOSH property, flattened (dot-style), for example `nats.password`. The key name is modified to be upper case and without dots in the context of an environment variable, therefore `nats.password` becomes `LINK_NATS_PASSWORD` in the container.
```
manno: So this is also wrong?
`nats.password`?

mook-as: No, as far as I know the `key` is still dot-separated.

manno:
I see, the change to 'dash' could be taken back, after implementing https://www.pivotaltracker.com/story/show/171433464

PR #832, opened 2020-02-26
Allow updates to versioned secret status, despite them being immutable

manno:
* versioned secret validator was in unrelated quarks secret package,
  quarks secret does not support the creation of versioned secret
* creating a new parent context in a request, does not create a new
  logger
* use existing request context
* add caller skip to logger, so source location is reported right

[#171328241](https://www.pivotaltracker.com/story/show/171328241)

### DIFF
--- a/pkg/kube/controllers/versionedsecret/validating_webhook.go
+++ b/pkg/kube/controllers/versionedsecret/validating_webhook.go
@@ -1,8 +1,10 @@
-package quarkssecret
+package versionedsecret
 
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"reflect"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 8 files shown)


pkg/kube/controllers/versionedsecret/validating_webhook.go
```diff
@@ -65,17 +66,17 @@ type ValidationHandler struct {
 
 // NewValidationHandler returns a new ValidationHandler
 func NewValidationHandler(log *zap.SugaredLogger) admission.Handler {
-	validationLog := log.Named("secret-validator")
+	validationLog := log.Named("secret-validator").Desugar().WithOptions(zap.AddCallerSkip(-1)).Sugar()
```
viovanov: fancy

manno: That's needed to make the code locations accurate... maybe we can clean that up (understand why -1) when doing https://www.pivotaltracker.com/story/show/170841351

Issue #812, opened 2020-02-06
Specifying a new readiness probe using an ops-file doesn't recreate the pod


f0rmiga:
On KubeCF, when I try to upgrade an existing installation to update a readiness probe, the pod doesn't get recreated.

edwardstudy:
I tried to reproduce this issue. And desired manifest has added the proper item: `quarks.run.healthcheck`. But quarks job `ig` didn't create new version of IG. Maybe it was equal to older one.

edwardstudy:
Opened a PR: https://github.com/cloudfoundry-incubator/cf-operator/pull/841

Hi, I examined the code changes thru this manifest to update:
https://github.com/cloudfoundry-incubator/cf-operator/blob/92f5724c205f4fc953cb4fb6d2da89e6fe81e86b/docs/examples/bosh-deployment/boshdeployment.yaml

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: nats-manifest
data:
  manifest: |
    ---
    name: nats-deployment
    releases:
    - name: nats
      version: "26"
      url: docker.io/cfcontainerization
      stemcell:
        os: opensuse-42.3
        version: 30.g9c91e77-30.80-7.0.0_257.gb97ced55
    instance_groups:
    - name: nats
      instances: 1
      jobs:
      - name: nats
        release: nats
        properties:
          nats:
            user: admin
            password: ((nats_password))
          quarks:
            ports:
            - name: "nats"
              protocol: "TCP"
              internal: 4222
            - name: "nats-routes"
              protocol: TCP
              internal: 4223
            run:
              healthcheck:
                nats:
                  readiness:
                    exec:
                      command:
                      - "echo healthy"
```

I could see quarksStatefulset changed about readinessProbe:
```
apiVersion: quarks.cloudfoundry.org/v1alpha1
kind: QuarksStatefulSet
metadata:
...
  name: nats-deployment-nats
  namespace: staging
...
spec:
  template:
    spec:
        spec:
          containers:
          - name: nats-nats
            readinessProbe:
              exec:
                command:
                - echo healthy
```

PR #841, opened 2020-03-03
Refine bpm for quarks properties

edwardstudy:
[#171141986](https://www.pivotaltracker.com/story/show/171141986)

fixes: #812 
- Added quarks properties to bpm configs
- ~~Got instance group from ig-resovled manifest~~
- Built containers properties('run', 'post_start', 'debug', 'activePassiveProbes') from bpm configs

And it should not break the current integration cases.

### DIFF
--- a/pkg/bosh/bpm/config.go
+++ b/pkg/bosh/bpm/config.go
@@ -62,12 +62,63 @@ type Port struct {
 // Config represent a BPM configuration
 type Config struct {
 	Processes           []Process `yaml:"processes,omitempty" json:"processes,omitempty"`
-	Ports               []Port    `yaml:"ports,omitempty" json:"ports,omitempty"`
 	UnsupportedTemplate bool      `json:"unsupported_template"`
+
+	// Bind these to reflect quarks properties update
+	Ports               []Port                  `yaml:"ports,omitempty" json:"ports,omitempty"`
+	Run                 RunConfig               `json:"run"`
+	PostStart           PostStart               `json:"post_start"`
+	Debug               bool                    `json:"debug"`
+	ActivePassiveProbes map[string]corev1.Probe `json:"activePassiveProbes"`
+}
+
+// RunConfig describes the runtime configuration for this job.
+type RunConfig struct {
+	HealthCheck     map[string]HealthCheck  `json:"healthcheck" yaml:"healthcheck"`
+	SecurityContext *corev1.SecurityContext `json:"security_context" yaml:"security_context"`
+}
+
+// HealthCheck defines liveness and readiness probes for a container.
+type HealthCheck struct {
+	ReadinessProbe *corev1.Probe `json:"readiness" yaml:"readiness"`
+	LivenessProbe  *corev1.Probe `json:"liveness"  yaml:"liveness"`
+}
+
+// PostStart allows post-start specifics to be passed through the manifest.
+type PostStart struct {
+	Condition *PostStartCondition `json:"condition,omitempty"`
+}
+
+// PostStartCondition represents the condition that should succeed in order to execute the
+// post-start script. It's often set to be the same as the readiness probe of a job.
+type PostStartCondition struct {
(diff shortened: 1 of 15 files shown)


viovanov:
@edwardstudy can you rebase please?

pkg/bosh/bpm/config.go
```diff
@@ -62,13 +62,80 @@ type Port struct {
...
+// PostStartCondition represents the condition that should succeed in order to execute the
+// post-start script. It's often set to be the same as the readiness probe of a job.
+type PostStartCondition struct {
+	Exec *corev1.ExecAction `json:"exec,omitempty"`
 }
 
 // Configs holds a collection of BPM configurations by their according job
 type Configs map[string]Config
 
+// IsActivePassiveModel indicates weather this bpm configs contains ActivePassiveProbes
```
viovanov: "whether these bpm configs contain"

pkg/kube/util/instancegroupmanifest/instance_group_manifest.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	bdm "code.cloudfoundry.org/cf-operator/pkg/bosh/manifest"
+	"code.cloudfoundry.org/quarks-utils/pkg/names"
+	"code.cloudfoundry.org/quarks-utils/pkg/versionedsecretstore"
+)
+
+// the key of instance group properties inside the secret
+var igPropertiesKey = "properties.yaml"
```
viovanov: capital T for consistency please

pkg/kube/util/instancegroupmanifest/instance_group_manifest.go
```diff
@@ -0,0 +1,53 @@
...
+
+// InstanceGroupManifest resolves references from desired manifest to a single instance group manifest
+type InstanceGroupManifest struct {
+	client               client.Client
+	versionedSecretStore versionedsecretstore.VersionedSecretStore
+}
+
+// NewInstanceGroupManifest constructs a resolver
```
viovanov: `constructs an InstanceGroupManifest `?

pkg/kube/controllers/watchnamespace/controller.go
```diff
@@ -16,11 +16,13 @@ import (
 	"code.cloudfoundry.org/quarks-utils/pkg/config"
 )
 
+// Reconcile represents reconcile.Reconciler
```
viovanov: `Reconcile implements reconcile.Reconciler`

pkg/kube/controllers/watchnamespace/controller.go
```diff
@@ -16,11 +16,13 @@ import (
...
 type Reconcile struct {
 }
 
 var _ reconcile.Reconciler = &Reconcile{}
 
+// Reconcile represents an empty reconciliation
```
viovanov: `// Reconcile doesn't do anything, because there is nothing to reconcile with the watched namespace`

edwardstudy:
@viovanov I wanted to notice that current PR is broken because we used one IG qJob to generate two secrets for ig-resovled properties and bpm configs. We could not guarantee which secret can be created firstly from same pod containers. 

As for that, if `bpm-configs-v2` was created before `ig-resovled properties-v2`, the bpm controller will get old `ig-resovled properties-v1` secret and get instances properties from this old one. This's a problem. Right?

That's why I asked to keep old procedure(get instance groups from desired manifest) and only added updated value to bpm config.

Does that make sense?

manno:
> That's why I asked to keep old procedure(get instance groups from desired manifest) and only added updated value to bpm config.

Yes, I think we talked about that in the daily? This means you'd just add all missing values to the BPM, right?

pkg/kube/util/instancegroupmanifest/instance_group_manifest.go
edwardstudy: Removed file

PR #843, opened 2020-03-05
Ig resolved secret should have az info

rohitsakala:
[#171152604](https://www.pivotaltracker.com/n/projects/2192232/stories/171152604)

### DIFF
--- a/pkg/kube/util/boshdns/dns.go
+++ b/pkg/kube/util/boshdns/dns.go
@@ -278,23 +278,15 @@ func (dns *boshDomainNameService) createCorefile(namespace string) (string, erro
 	for _, alias := range dns.Aliases {
 		for _, target := range alias.Targets {
 			// Implement BOSH DNS placeholder alias: https://bosh.io/docs/dns/#placeholder-alias.
-			if target.Query == "_" {
-				instanceGroup, found := dns.InstanceGroups.InstanceGroupByName(target.InstanceGroup)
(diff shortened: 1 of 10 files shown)


pkg/kube/util/boshdns/dns.go
```diff
@@ -283,12 +283,24 @@ func (dns *boshDomainNameService) createCorefile(namespace string) (string, erro
...
 				}
-				for i := 0; i < instanceGroup.Instances; i++ {
-					id := fmt.Sprintf("%s-%d", target.InstanceGroup, i)
-					from := strings.Replace(alias.Domain, "_", id, 1)
-					serviceName := instanceGroup.IndexedServiceName(dns.ManifestName, i)
-					to := fmt.Sprintf("%s.%s.svc.%s", serviceName, namespace, clusterDomain)
-					rewrites = append(rewrites, dnsTemplate(from, to, target.Query))
+				if len(instanceGroup.AZs) > 0 {
+					for azIndex := range instanceGroup.AZs {
+						for i := 0; i < instanceGroup.Instances; i++ {
```
manno: This is too deeply nested for my taste: for,for,if,if,for,for. Maybe some of that code can go into a separate function?

There is also some overlap with the else case below.

rohitsakala: True, Thanks for the comment. Hope, I fixed it now

PR #852, opened 2020-03-16
Ability to use PVCs for ephemeral disks

viovanov:
## Description
Flag to use a PVC for ephemeral disks.

## Motivation and Context
Required in cases where the ephemeral mount is assumes to be exclusively used by a job.

## How Has This Been Tested?
Locally

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [ ] needs docs explaining the flags
- [ ] needs tests


pkg/bosh/bpmconverter/volume_factory.go
```diff
@@ -358,29 +444,18 @@ func resolvedPropertiesVolumeMount(name string, instanceGroupName string) corev1
...
-}
-
 func sysDirVolume() *corev1.Volume {
 	return &corev1.Volume{
 		Name:         VolumeSysDirName,
 		VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
 	}
 }
 
+// VolumeDataDirName is the volume name for the data directory.
```
manno: Does this have to be public? I can't find any usage  outside `bpmconverter`.
In the tests I would prefer to use the constant string result. Feels more decoupled from the implementation. Also  eliminates the need for `VolumeDataDirName` unit tests.

Actually, if this is only used in bpmconverter, the func can be replaced with `generateEphemeralPersistentVolumeClaimName` completely.

pkg/bosh/bpmconverter/volume_factory.go
```diff
@@ -307,6 +389,10 @@ func generatePersistentVolumeClaim(manifestName string, instanceGroup *bdm.Insta
 	return persistentVolumeClaim
 }
 
+func generateEphemeralPersistentVolumeClaimName(manifestName string, instanceGroupName string) string {
```
manno: This is a long func name. I feel we use 'generate' a bit too much. Maybe `ephemeralName` is precise enough?

pkg/bosh/bpmconverter/volume_factory.go
```diff
@@ -282,6 +328,42 @@ func (f *VolumeFactoryImpl) GenerateBPMDisks(manifestName string, instanceGroup
 	return bpmDisks, nil
 }
 
+func generateEphemeralPersistentVolumeClaim(manifestName string, instanceGroup *bdm.InstanceGroup, namespace string) corev1.PersistentVolumeClaim {
+	// Spec of a persistentVolumeClaim
```
manno: This comment followed by a newline seems unrelated? Maybe this is better as a func comment?

pkg/bosh/bpmconverter/volume_factory.go
```diff
@@ -282,6 +328,42 @@ func (f *VolumeFactoryImpl) GenerateBPMDisks(manifestName string, instanceGroup
...
+
+	diskSize := defaultEphemeralVolumeSize
```
manno: Just nitpicking: generating the disk size could be a separate private function. I don't mind keeping it inline, but it looks like it really belongs to the instance group struct and we tend to amass these procedural, intervined if blocks in our 'generate' funcs :)

pkg/bosh/bpmconverter/volume_factory.go
```diff
@@ -243,6 +263,32 @@ func (f *VolumeFactoryImpl) GenerateBPMDisks(manifestName string, instanceGroup
 					"ephemeral": "true",
 				},
 			}
+
+			if persistent {
```
manno: I think the whole func is to deeply nested. If you want, this is an opportunity to extract some blocks to methods.

```
if hasEphemeralDisk {
  disks := append(disks, newEphemeralDisk(...))
}
```

pkg/bosh/bpmconverter/volume_factory.go
viovanov: going with `ephemeralPVCName`

PR #853, opened 2020-03-17
Run cf-operator as numeric user

gdankov:
## Description
In order to run cf-operator right now, you need a permissive PSP in the cluster which allows it to run as root. Even though you are running it as `vcap`, Kubernetes PSPs can only validate against numeric user ids.

## Motivation and Context
cf-operator cannot be run without providing an additional permissive PSP. This can be easily avoided with this change. 

## How Has This Been Tested?
It hasn't been tested with cf-operator, but it should work because it's a Kubernetes feature anyway.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project.

cc @herrjulz


gdankov:
Created a draft PR by mistake 🙈

PR #856, opened 2020-03-20
Trace logger for addons

manno:
[#17084135](https://www.pivotaltracker.com/story/show/170841351)


viovanov:
@manno I think lint is failing

PR #860, opened 2020-03-24
Add helm requirements.yaml for quarks-job

edwardstudy:
[#171492037](https://www.pivotaltracker.com/story/show/171492037)

- Added helm requirements.yaml for quarks-job
- Bumped quarks-job to 491e84a


deploy/helm/cf-operator/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: quarks-job
+    version: 0.0.0+0.g06ad02b
```
manno: I'd prefer it, if we had to change the version only in one file. Can we use `$QUARKS_JOB_HELM_VERSION` here?
Also this says 06ad02b and bin/include/dependencies says 491e84a

manno: I missed the replace script

edwardstudy: I used `0.0.0` as placeholder.

edwardstudy:
Rebased and verified with helm v3.1.2

```
./bin/build-helm
Successfully packaged chart and saved it to: /Users/zidashaw/space/golang/workspace/src/github.com/cloudfoundry-incubator/cf-operator/helm/cf-operator-4.0.0+dirty-10.g5a91f8bf.tgz
```
image

PR #882, opened 2020-04-08
Include a list of referenced images in the helm chart

aduffeck:
This pull requests adds a list of references images to the helm chart. That helps with tooling consuming this chart and its images.


bin/build-helm
```diff
@@ -22,8 +22,29 @@ perl -pi -e "s|org: .*|org: ${DOCKER_IMAGE_ORG}|g" "${output_dir}/cf-operator/va
...
+  puts values['operator']['boshDNSDockerImage']
+
+  values = YAML.load_file('${quarks_job_values_file}')
+  image = values['image']
+  puts "#{image['org']}/#{image['repository']}:#{image['tag']}"
+EOF
+rm "${quarks_job_values_file}"
+
 helm package helm/cf-operator \
-  --dependency-update \
```
viovanov, line 26: was this intentional?

aduffeck: Yes. We need to download the qj helm chart explicitly before packaging the operator chart so that we can inspect its `values.yaml` and get the referenced images from it.

viovanov: makes sense

PR #901, opened 2020-04-21
fix: Permissions and ownership for copied secrets

viovanov:
## Description
This change allows the operator to have a cache-enabled client that can span multiple namespaces.
Also fixes a bug with cross-namespace ownership.

## Motivation and Context
Required for copies of generated secrets to work.

## How Has This Been Tested?
With e2e tests and KubeCF.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code has security implications.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/deploy/helm/cf-operator/templates/cluster-role.yaml
+++ b/deploy/helm/cf-operator/templates/cluster-role.yaml
@@ -17,6 +17,14 @@ rules:
   - watch
 
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - update
+
+- apiGroups:
   - certificates.k8s.io
   resources:
   - certificatesigningrequests/approval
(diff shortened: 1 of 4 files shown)


deploy/helm/cf-operator/templates/cluster-role.yaml
```diff
@@ -43,6 +43,18 @@ rules:
   - update
   - watch
 
+- apiGroups:
```
manno, line 46: the secrets block is duplicated

pkg/kube/controllers/quarkssecret/quarkssecret_reconciler.go
```diff
@@ -417,13 +488,35 @@ func (r *ReconcileQuarksSecret) createSecret(ctx context.Context, qsec *qsv1a1.Q
 
 	secret.SetLabels(secretLabels)
 
+	// If this is a copy (lives in a different namespace), we only do an update,
```
manno, line 491: I'd prefer it if this block could live somewhere else, because the `createSecret` func is called from several places like a generic helper, but now has some added functionality to do something else entirely.

The if block doesn't seem to need the labels from line 489, maybe `createSecrets` could  use a  specialized `createSecretCopies` instead?

Issue #891, opened 2020-04-13
CSR permissions error on Kubernetes 1.18 with RBAC enabled


curtailrr:
**Describe the bug**
Deploying KubeCF 1.0.1 with cf-operator 3.3.0.
After pod/kubecf-database-0 starts, csr/kubecf-kubecfvar-bits-ser is stuck in the Pending state.

Looking at cf-operator logs:
ERROR	csr-reconciler	quarkssecret/certificatesigningrequest_reconciler.go:167	Failed to approve certificate signing request: could not update approval of CSR ‘kubecf-kubecfvar-bits-ser’: certificatesigningrequests.certificates.k8s.io “kubecf-kubecfvar-bits-ser” is forbidden: user not permitted to approve requests with signerName “kubernetes.io/legacy-unknown”

I was able to work around this issue with:
kubectl edit -n cf-operator clusterrole cf-operator

I added:

- apiGroups:
  - certificates.k8s.io
  resourceNames:
  - kubernetes.io/legacy-unknown
  resources:
  - signers
  verbs:
  - approve

**To Reproduce**
Install microk8s with Kubernetes 1.18, enable RBAC, then install cf-operator 3.3.0 followed by KubeCF 1.0.1.

**Expected behavior**
CSR should be approved, deployment should proceed.

**Environment**
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
cf-operator	cf-operator	1       	2020-04-13 23:05:36.654924578 +0000 UTC	deployed	cf-operator-3.3.0+0.gf32b521e	3.3.0+0.gf32b521e
kubecf	kubecf   	1       	2020-04-13 23:06:40.586228721 +0000 UTC	deployed	kubecf-v1.0.1	v1.0.1     

**Additional context**

curtailrr:
[...]

PR #914, opened 2020-04-30
Add testing of kubernetes 1.17.0,1.18.0 in travis

rohitsakala:
[#172397180](https://www.pivotaltracker.com/n/projects/2192232/stories/172397180)


manno:
1.18 is not green, https://github.com/cloudfoundry-incubator/cf-operator/issues/891

.travis.yml
```diff
@@ -20,6 +20,8 @@ env:
     - KUBE=1.14.10
     - KUBE=1.15.7
     - KUBE=1.16.4
+    - KUBE=1.17.0
+    - KUBE=1.18.0
```
manno, line 24: I'm concerned this will make our test matrix to large. If travis doesn't schedule enough workers, tests are executed sequentially with a wait time.

mudler: can we drop 1.14.x ? or we might see how it behaves and come back to it

rohitsakala: I will block this PR until we create a story for the nightly travis. It is taking too much time already.

PR #922, opened 2020-05-06
165900520 multinamespace

manno:
React on events in all labeled namespaces ([README](https://github.com/cloudfoundry-incubator/cf-operator/blob/ab21f40a62a31fae75088edc709a787c9691fb0b/deploy/helm/cf-operator/README.md#using-multiple-operators))

[#165900520](https://www.pivotaltracker.com/story/show/165900520)


bin/dev-helm
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+namespace=${CF_OPERATOR_NAMESPACE-cfo}
+kubecf_location=${KUBECF_LOCATION-~/Downloads/kubecf-v2.1.0.tgz}
+
+case "$1" in
```
mudler, line 6: :heart: was about to add something similar too

deploy/helm/cf-operator/templates/single-namespace.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.global.singleNamespace.create }}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "{{ .Values.global.singleNamespace.name }}"
+  labels:
+    quarks.cloudfoundry.org/monitored: {{ .Values.global.monitoredID }}
```
mudler, line 8: :+1: in this way we don't set up any label during startup.. much nicer

PR #937, opened 2020-05-11
Fix bug with multiple versions when updating

manno:
This should fix the tests.

* GetMaxStatefulSetVersion was called with an empty namespace.
* The webhooks scoping was not taking effect, because the scope was still
set to '*'
* Also fixes logged code location in webhooks

[#165900520](https://www.pivotaltracker.com/story/show/165900520)


pkg/kube/controllers/boshdeployment/bpm_reconciler.go
```diff
@@ -197,14 +197,18 @@ func (r *ReconcileBPM) applyBPMResources(bdplName string, instanceGroupName stri
...
+
+	qStsVersionString := "0"
+	if quarksStatefulSet.Namespace != "" {
+		_, qStsVersion, err := qstscontroller.GetMaxStatefulSetVersion(r.ctx, r.client, quarksStatefulSet)
+		if err != nil {
+			return nil, err
+		}
+		qStsVersion = qStsVersion + 1
+		qStsVersionString = strconv.Itoa(qStsVersion)
+		if err != nil {
```
mudler, line 209: Superflous check here? it's still the same `err` scoped here

PR #942, opened 2020-05-14
Consume quarks-utils common scripts

mudler:
## Description
Use a wrapper to checkout quarks-utils scripts pinned in `go.mod` and consume them for common tasks

Depends on: https://github.com/cloudfoundry-incubator/quarks-utils/pull/48

## Motivation and Context

[#172645579](https://www.pivotaltracker.com/story/show/172645579)

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.


.travis.yml
```diff
@@ -73,10 +71,8 @@ jobs:
         - go get -u golang.org/x/lint/golint
       name: lint
       script:
-        - bin/lint
-        - bin/vet
-        - bin/staticcheck
```
manno, line 78: This looks like staticcheck has been removed?

Makefile
```diff
@@ -11,17 +14,21 @@ up:
 vet:
 	bin/vet
 
-lint:
-	bin/lint
+lint: tools
+	$(QUARKS_UTILS)/bin/lint
 
+.PHONY: tools
```
manno, line 20: In contrast to most of the other targets in this Makefile, I think this one not phony, meaning it should only run if the dir is missing.

.travis.yml
mudler: It has been moved as part of the lint process (as it was in quarks-jobs) https://github.com/cloudfoundry-incubator/quarks-utils/blob/master/bin/lint#L5.  I would go over to drop the separate job in the pipeline in case we merge this

Makefile
mudler: I've made it phony because if you don't run it again, you loose the check if you are up-to-date with quarks-utils (which is what `bin/tools` does). 

The bin/tools will sync the quarks-utils dependency even if you have already checked it out.

I've give it a thought at it, and the issue would be that if you change the go.mod to point to a different version of quarks-utils,  you might have a false-sign that you are up-to-sync, while a manual intervention is required

.travis.yml
manno: Not sure this has the same fail behavior as lint. Go lint is not meant for failing PRs (https://github.com/golang/lint#purpose). 

Also I suspect vet and lint need to download modules to work, but staticcheck does not. On concourse we would do less work if vet and lint were running in one task.

Makefile
manno: I read the bin/tools script too late, sorry :)

I'll approve this PR so we can move to the new scripts. But I have huge concerns that this is too much magic, not due this PR, but for the story.

.travis.yml
mudler: :+1:  I can amend that and run staticcheck only in travis here as I've left the target in the makefile, but the fix can go in another PR for quarks-utils as the common scripts are already there

.travis.yml
```diff
@@ -56,11 +56,9 @@ before_script:
 script:
   - bin/build-image
   - kind load docker-image "cfcontainerization/cf-operator:dev" --name kind$KUBE
-  - bin/test-integration
-  - bin/test-integration storage
-  - bin/test-integration util
-  - bin/test-cli-e2e
-  - bin/test-helm-e2e
+  - make test-integration test-integration-storage test-integration-subcmds
```
manno, line 59: I'd prefer using the bin scripts directly, in travis there is no need to check for dependencies every time

.travis.yml
mudler: I've just checked, we were already making it fail with `bin/lint`, should we change behavior here ? I don't have strong feelings on this

.travis.yml
mudler: Fixed in https://github.com/cloudfoundry-incubator/quarks-operator/pull/942/commits/99af96c7783f50d1b4d5801e595a96987e956c28 :+1:

Makefile
mudler: No worries, yeah I agree.. this approach ended up being very much magical, and I already start to feel that will hit us later _iff_ the projects will start to diverge...

.travis.yml
manno: I remember arguing about that before, but let's just keep it for now

PR #956, opened 2020-05-26
Grep the quarks-job docker image version out of go.mod

manno:
Not sure if this will work with dev versions.


manno:
This will probably be flaky until https://github.com/cloudfoundry-incubator/quarks-operator/pull/960 is merged

PR #964, opened 2020-05-29
Add docs website

mudler:
## Description
Re-structure docs and add GH action workflow to push the generated content (with hugo) to https://quarks.suse.dev . There is still room for improvement and refactoring of docs, but I think it's a good start

## Motivation and Context
[172598851](https://www.pivotaltracker.com/story/show/172598851)

## How Has This Been Tested?
https://quarks.suse.dev

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.


mudler:
aand I should have added a bit more on details on that, doing it now for reference:

For local development, the only useful target is `make website-serve`, there the only dependency needed  it's `hugo` which will be fetched automatically at the pinned version if not already present (it should also work on macos, I got positive feedback from @fargozhu that tested it on kubecf-docs IIRC). 
The `npm` dependency you see in the scripts it's just used when building the website docs for production (so only CI will run it).

I've also setted up a GH action which will build and automatically publish a new version of the website only if we touch the website code, the docs/ folder, or the pipeline itself.

PR #998, opened 2020-06-24
Entanglement improvements and tests

manno:
[#172993878](https://www.pivotaltracker.com/story/show/172993878)

## Checklist:

Check item if necessary.

- [x] Review PRs of changed Quarks dependencies
- [x] Update this PR after the release of Quarks dependencies


pkg/kube/controllers/boshdeployment/deployment_controller.go
```diff
@@ -85,7 +86,7 @@ func AddDeployment(ctx context.Context, config *config.Config, mgr manager.Manag
 	}
 
 	// Watch ConfigMaps referenced by the BOSHDeployment
-	configMapPredicates := predicate.Funcs{
```
edwardstudy, line 88: As for readability, why we need to shorten this variable and reuse it?

manno: Mostly because the other controllers also do that and for having shorter variable names: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

PR #1004, opened 2020-06-29
Simpler tools-shared script to download shared tools

manno:
[#173008965](https://www.pivotaltracker.com/story/show/173008965)


bin/tools-shared
```diff
@@ -1,34 +1,2 @@
...
-    git fetch --all
-    if [[ $(git --no-pager diff HEAD "$quarks_utils_version" --raw | wc -l) != "0" ]]; then
-      echo "Changes detected, resetting branch"
-      git reset --hard "$quarks_utils_version"
-    fi
-  fi
-
-popd
+#!/bin/sh
+curl -s "https://raw.githubusercontent.com/cloudfoundry-incubator/quarks-utils/master/bin/tools-shared-get-v1.sh" | sh
```
mook-as, line 2: I feel *very* strongly that we should not `curl | sh` with no validity checks:
- This won't work offline (e.g. in OBS builds)
- You're going to end up executing random HTTP 500s (complete with shell redirection, because `<html>`).
- This is completely unreproducible (because it's `master`, we have no clue what worked five years ago).
  - This also means if multiple repos depend on quarks-utils, things are guaranteed to be broken as the API changes (because `master`).
- And that's all completely ignoring potential security issues.

I know git submodules are a pain, but at least they're better than _this_.  Or we can do GitHub releases / tarballs with a checksum verification.  Or vendor the whole thing.  Even package everything in RPMs or something.

manno: Hey @mook-as, why would you run this in OBS? 
We use this in Github Actions, couldn't you use another approach if you're running in OBS?

Also, this is just the loader, not sure if you looked at https://raw.githubusercontent.com/cloudfoundry-incubator/quarks-utils/master/bin/tools-shared-get-v1.sh - it does use a Github tarball.

As for the concerns regarding master, the script has a version in it's filename. I feel like that's already over-engineering.
Please note that we also get other dependencies like that, i.e. in an unreproducible way, without checksums:
https://github.com/cloudfoundry-incubator/quarks-operator/blob/master/.github/workflows/ci.yaml#L27
Normally CI just goes red and we fix it.

mudler: I have mixed feelings about this approach, I'm not a fan of it but I wouldn't block from merging this because it does indeed satisfies our requirements, which aren't very strict. 

Those scripts serves purely for development, and from a package management standpoint they are not part of the toolchain as quarks-operator should be built as a standard golang project. So in the OBS case, If you intend to use the shared script, you can get it separately as part of the packaging process, but you should be able to `go build` the project. The usage of `go mod` without shipping vendored modules in the source code already breaks airgapped builds.

I'm totally all in for a PACKAGING guideline, but that being said, if we have issues to just `go build` the project, then we should solve that first.

PR #1013, opened 2020-07-02
Change ExpectAllKeys to True

rohitsakala:
[#173329326](https://www.pivotaltracker.com/n/projects/2192232/stories/173329326)

This is required because only then the variable interpolation job will fail if every variable is not rendered in the desired-manifest

### DIFF
--- a/pkg/bosh/manifest/cmd_interpolate.go
+++ b/pkg/bosh/manifest/cmd_interpolate.go
@@ -61,7 +61,7 @@ func InterpolateFromSecretMounts(boshManifestBytes []byte, variablesDir string, 
 		}
 	}
 
-	yamlBytes, err := InterpolateExplicitVariables(boshManifestBytes, vars)
+	yamlBytes, err := InterpolateExplicitVariables(boshManifestBytes, vars, true)
 	if err != nil {
 		return errors.Wrap(err, "failed to interpolate explicit variables")
 	}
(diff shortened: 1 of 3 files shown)


rohitsakala:
> Could we add one or two test cases for this option `ExpectAllKeys`?

Done

integration/bdpl_update_test.go
```diff
@@ -348,20 +348,27 @@ var _ = Describe("BDPL updates", func() {
...
 				Expect(err).NotTo(HaveOccurred(), "error updating user var")
 				tearDowns = append(tearDowns, tearDown)
 
 				err = env.WaitForSecret(env.Namespace, "desired-manifest-v3")
-				Expect(err).NotTo(HaveOccurred(), "error waiting for new desired manifest")
+				Expect(err).NotTo(HaveOccurred(), "error waiting for new desired manifest v3")
 
-				secret, err := env.GetSecret(env.Namespace, "desired-manifest-v3")
-				Expect(err).NotTo(HaveOccurred(), "error getting new desired manifest")
+				err = env.CustomWaitForSecret(env.Namespace, "desired-manifest-v4", 20*time.Second)
```
edwardstudy, line 360: Could we get the latest versioned secret for this case?

rohitsakala: Yeah. Makes sense. Let me update that.

rohitsakala: Removed that change. Will make another PR is necessary.

PR #1018, opened 2020-07-06
Add e2e multiple namespace test

mudler:
Adds boshdeployment multi-namespace test

## Motivation and Context

Requires https://github.com/cloudfoundry-incubator/quarks-utils/pull/62

[#172551603](https://www.pivotaltracker.com/story/show/172551603)

## Checklist:

Check item if necessary.

- [x] Review PRs of changed Quarks dependencies

### DIFF
--- a/docs/examples/bosh-deployment/quarks-gora.yaml
+++ b/docs/examples/bosh-deployment/quarks-gora.yaml
@@ -1,0 +1,53 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ops-scale
+data:
+  ops: |
+    - type: replace
+      path: /instance_groups/name=quarks-gora?/instances
+      value: 2
+---
+apiVersion: v1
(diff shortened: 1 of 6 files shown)


mudler:
:thinking: mmh to be different from the current bdpl_update it needs to be a bit more complex. maybe as part of another story? wdyt @manno ?

manno:
@mudler  Yes, I think that should be part of the story. It's sufficient to just increase the instance count or add a variable like in https://github.com/cloudfoundry-incubator/quarks-operator/blob/master/integration/bdpl_update_test.go

mudler:
Now depends on https://github.com/cloudfoundry-incubator/quarks-utils/pull/66

mudler:
@rohitsakala @manno  addressed in https://github.com/cloudfoundry-incubator/quarks-operator/pull/1018/commits/c12e8370b025e71c114cda178e7db57b3dbb3689

PR #1053, opened 2020-08-07
cmd fixes

manno:
Cleanup and a fix for a possible silent error when experiencing IO problems.

[#174227363](https://www.pivotaltracker.com/story/show/174227363)


cmd/internal/varinterpolation.go
```diff
@@ -30,6 +30,7 @@ interpolated manifest to STDOUT
 	PreRun: func(cmd *cobra.Command, args []string) {
 		boshManifestFlagViperBind(cmd.Flags())
 		outputFilePathFlagViperBind(cmd.Flags())
+		viper.BindPFlag("variables-dir", cmd.Flags().Lookup("variables-dir"))
```
edwardstudy, line 33: Something like?
```
if err := viper.BindPFlag("variables-dir", cmd.Flags().Lookup("variables-dir")); err != nil {
    log.Fatalf("Unable to add variables-dir flag:%v", err)
}
```

manno: I'd ignore that linter warning. If we want to check BindPFlag, we should check all of them, not just one. Sadly we don't currently...

PR #1054, opened 2020-08-07
Add logrotate shell-out to tail-logs command

manno:
[#174227363](https://www.pivotaltracker.com/story/show/174227363)

### DIFF
--- a/cmd/internal/tlogs.go
+++ b/cmd/internal/tlogs.go
@@ -12,7 +12,9 @@ import (
 	"github.com/hpcloud/tail"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"go.uber.org/zap"
 
+	"code.cloudfoundry.org/quarks-operator/pkg/kube/util/logrotate"
 	"code.cloudfoundry.org/quarks-utils/pkg/cmd"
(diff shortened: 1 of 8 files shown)


cmd/internal/tlogs.go
```diff
@@ -26,24 +28,53 @@ var tailLogsCmd = &cobra.Command{
...
 The dir can be set using the "-z" flag, or setting
 the LOGS_DIR env variable.
+It will also run logrotate.
 
 `,
-	RunE: func(cmd *cobra.Command, args []string) (err error) {
-		return TailLogsFromDir()
+	RunE: func(_ *cobra.Command, args []string) (err error) {
+		log = logger.New(cmd.LogLevel())
+		defer log.Sync()
```
manno, line 36: was not checked before

cmd/internal/tlogs.go
```diff
@@ -26,24 +28,53 @@ var tailLogsCmd = &cobra.Command{
...
+
+		d := time.Duration(viper.GetInt("logrotate-interval")) * time.Minute
+		for range time.Tick(d) {
```
manno, line 39: I believe this is fine, since Run is essentially like main

PR #1055, opened 2020-08-07
Adapt manual command generation target

mudler:


Makefile
```diff
@@ -83,8 +83,7 @@ gen-fakes:
 	bin/gen-fakes
 
 gen-command-docs:
-	rm -f docs/commands/*
```
manno, line 86: Without the delete we will keep old files around?

mudler: yup.. but maybe should we put in gitignore instead? I don't think we will use this target much in this repository, hopefully we can just deprecate it

PR #1063, opened 2020-08-11
Support CPU limits

mudler:
Adds support for CPU limits in the BPM converter.

It also adds e2e test and enhance unit test to cover the new feature

~Will follow-up a separate PR for docs updates~ https://github.com/cloudfoundry-incubator/quarks-docs/pull/2

## Motivation and Context

[#174207985](https://www.pivotaltracker.com/story/show/174207985)

## Checklist:

Check item if necessary.

- [ ] Review PRs of changed Quarks dependencies


e2e/kube/cpu_mem_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"path"
+
+	cmdHelper "code.cloudfoundry.org/quarks-utils/testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("BOSHDeployment", func() {
+	When("specifying cpu/memory resources requests and limits in bpm configs", func() {
```
manno, line 13:
```suggestion
	When("specifying cpu/memory resources requests and limits in the quarks.bpm config", func() {
```

Something like that, I'd try to make it clear that this our config, not a BOSH one. There might be a test in the future, that checks if we convert existing "BOSH bpm limit section" information correctly.

mudler: Right! sounds fair :+1:

PR #1072, opened 2020-08-14
use manifest interpolation from bosh-cli

manno:
[#171895821](https://www.pivotaltracker.com/story/show/171895821)


manno:
Tests are green, even though https://github.com/cloudfoundry/bosh-cli/blob/master/director/template/template.go#L97-L112 doesn't type all keys as strings.

PR #1074, opened 2020-08-17
Remove container-run-copier

manno:
[#174357409](https://www.pivotaltracker.com/story/show/174357409)


manno:
We keep container-run in the operator so we can bump it independently from the stemcell.

PR #1078, opened 2020-08-19
Add BOSH DNS handler support

manno:
[#173814835](https://www.pivotaltracker.com/story/show/173814835)


manno:
Need to use multiple server blocks instead: https://coredns.io/manual/toc/#server-blocks

rohitsakala (approved):

PR #1103, opened 2020-08-27
Fix invalid BOSH DNS type error

manno:
Using `Type: http` would have resulted in an error:

```
                      Err: {
                          s: "template: Corefile:4:104: executing \"Corefile\" at <.Source.Type.HTTP>: can't evaluate field HTTP in type string",
                      },
```

Also adds support for all coredns server protocols.

https://coredns.io/manual/toc/#specifying-a-protocol

[#172994188](https://www.pivotaltracker.com/story/show/172994188)

### DIFF
--- a/pkg/kube/util/boshdns/corefile.go
+++ b/pkg/kube/util/boshdns/corefile.go
@@ -26,6 +26,7 @@ type Alias struct {
 }
 
 // Handler redirects DNS queries for a zone to a forward server
+// https://coredns.io/plugins/forward/
 type Handler struct {
 	Domain string        `json:"domain"`
 	Source HandlerSource `json:"source"`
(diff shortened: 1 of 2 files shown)


pkg/kube/util/boshdns/corefile.go
```diff
@@ -42,14 +42,18 @@ type HandlerSource struct {
...
-	return h.Type == "http"
-}
-
-// DNS returns true if the handler's source is of type dns
-func (h HandlerSource) DNS() bool {
-	return h.Type == "dns"
+// Protocol returns the coredns server protocol.
+func (h HandlerSource) Protocol() string {
+	switch h.Type {
+	case "http", "tls":
```
manno, line 48: I thought 'http' for 'tls' appeared in the original story and was used by BOSH somehow, but can't find it anymore. Maybe we should drop this, as it is confusing and just support the Coredns settings (dns, tls, https, grpc).

viovanov (approved):

PR #1105, opened 2020-08-31
Fix IGs so they don't need to restart when scaling

viovanov:


cmd/internal/trender.go
```diff
@@ -58,7 +58,24 @@ This will render a provided manifest instance-group
 
 		replicas := viper.GetInt("replicas")
 		if replicas < 0 {
-			return errors.Errorf("%s replicas flag is empty.", tRenderFailedMessage)
+			replicas = 1
```
f0rmiga, line 61: This sounds weird. Could you leave a comment for future reference why this is needed?

cmd/internal/trender.go
```diff
@@ -58,7 +58,24 @@ This will render a provided manifest instance-group
...
+		}
+
+		podOrdinal := viper.GetInt("pod-ordinal")
+		if podOrdinal < 0 {
```
f0rmiga, line 65: Same thing here, why would it be < 0 ever?

cmd/internal/trender.go
```diff
@@ -69,28 +86,17 @@ This will render a provided manifest instance-group
...
-					return errors.Errorf("%s Cannot extract podOrdinal flag value from hostname %s", tRenderFailedMessage, hostname)
-				}
-				podOrdinal, err = strconv.Atoi(match[1])
-				if err != nil {
-					return errors.Wrapf(err, "%s String to int conversion failed from hostname for calculating podOrdinal flag value %s", tRenderFailedMessage, hostname)
-				}
-			}
-			specIndex = (azIndex-1)*replicas + podOrdinal
+
+			specIndex = (azIndex-1)*10000 + podOrdinal
```
f0rmiga, line 90: `10000` is magical here, could you make a const out of it to be more readable? Alternatively, stick a comment to it.

f0rmiga (approved):

PR #1153, opened 2020-09-28
Implicit variables and var/key notation?

manno:
https://www.pivotaltracker.com/story/show/174969390

Trying to understand implicit variables, slash vs dot.


manno:
I think this is a special workaround we added.

PR #1154, opened 2020-09-28
Add release-drafter to Github workflow

manno:

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,22 @@
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'feature'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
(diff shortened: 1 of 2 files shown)


.github/release-drafter.yml
```diff
@@ -0,0 +1,22 @@
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'feature'
```
rohitsakala, line 4: should we add `feat` also ? I see it in the documentation.

rohitsakala (approved):

.github/release-drafter.yml
manno: Hm, we only have an "enhancement" label, also tracker seems to use that, too. I copied most of that config from the docs, so we might be able to drop 'feature', too. But I would wait and see if maybe that's there for a reason....

rohitsakala: :+1:

PR #1156, opened 2020-09-29
Refactor operator upgrade tests

manno:
[#174250908](https://www.pivotaltracker.com/story/show/174250908)

### DIFF
--- a/pkg/kube/controllers/boshdeployment/bpm_reconciler.go
+++ b/pkg/kube/controllers/boshdeployment/bpm_reconciler.go
@@ -3,6 +3,7 @@ package boshdeployment
 import (
 	"context"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


pkg/kube/controllers/boshdeployment/bpm_reconciler.go
```diff
@@ -138,8 +139,11 @@ func (r *ReconcileBPM) Reconcile(request reconcile.Request) (reconcile.Result, e
 	// Apply BPM information
 	resources, err := r.applyBPMResources(bdpl.Name, instanceGroupName, bpmSecret, manifest, dnsService.Spec.ClusterIP)
 	if err != nil {
-		return reconcile.Result{},
-			log.WithEvent(bpmSecret, "BPMApplyingError").Errorf(ctx, "Failed to apply BPM information: %v", err)
+		if apierrors.IsNotFound(err) {
+			log.WithEvent(bpmSecret, "SkipReconcile").Debugf(ctx, "Requeue reconcile: %s", err)
+			return reconcile.Result{RequeueAfter: time.Second * 5}, nil
```
rohitsakala, line 144: If the `ig-resolved-v1` secret is not created due to an error in the QuarksJob, then this will go in an infinite loop. Is it better to leave it to the controller runtime implementation of  reconcile (err != nil) logic here ?

rohitsakala: +1. Had a call.

rohitsakala (approved):

PR #1162, opened 2020-10-01
Remove regex for * query type in cname template

rohitsakala:
[#174694400](https://www.pivotaltracker.com/n/projects/2192232/stories/174694400)

### DIFF
--- a/e2e/kube/examples_test.go
+++ b/e2e/kube/examples_test.go
@@ -368,7 +368,7 @@ var _ = Describe("Examples Directory", func() {
 			}
 		})
 
-		It("resolves BOSH DNS wildcard aliases", func() {
+		It("resolves BOSH DNS aliases", func() {
 			By("Getting expected IP")
 			podName := "nats-0"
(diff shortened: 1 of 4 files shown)


mudler (approved):

integration/bdpl_update_test.go
```diff
@@ -290,7 +290,7 @@ var _ = Describe("BDPL updates", func() {
 			It("should update the dns pods", func() {
 				deployment, err := env.CollectDeployment(env.Namespace, "bosh-dns", 1)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(deployment.Spec.Template.GetAnnotations()).To(HaveKey(quarksrestart.RestartKey))
+				Expect(deployment.Spec.Template.GetAnnotations()).To(HaveKey(quarksrestart.AnnotationRestartOnUpdate))
```
manno, line 293: Does this change the meaning of the test? Like from "has restarted" to "should restart"?

rohitsakala: Nice catch. This was a mistake.

PR #1164, opened 2020-10-05
Change the coredns image source and tag

manno:
There is a problem with the CAP release tool, as it can't handle tags.
However we only get a SHA from OBS: https://jira.suse.com/browse/CAP-1598

Additionally, we're told registry.opensuse.org cannot store multiple
versions, so the SHA would disappear once a new version is pushed. All
registry.opensuse.org images have to be copied elsewhere.

Switching back to dockerhub and tag for now.


dragonchaser (approved):
lgtm

PR #1181, opened 2020-10-08
fix: pre-upgrade hook args

f0rmiga:
https://github.com/SUSE/kubecf-kubectl/blob/3f446170f0e8147967a1e8858309754e76e63a91/Dockerfile#L66


mook-as (approved):
Approving since the rest of the code is already there, it's just to match the changed image.

That means I can short circuit thinking about if that's the correct style and everything :D

PR #1209, opened 2020-10-27
Upload the helm chart for successful GH workflows

manno:
[#174254515](https://www.pivotaltracker.com/story/show/174254515)


rohitsakala (approved):
So, from now on, we will be providing the links of github action artifacts.

manno:
> So, from now on, we will be providing the links of github action artifacts.

Not sure if anyone is using that. This is only for providing the PR results.

PR #1220, opened 2020-10-30
Add support for JSON string in implicit vars

manno:
As discussed, this should add support for JSON to implicit secrets.

The implicit variable secret needs to have the JSON annotation.


jandubois:
Is there any recommended value for the annotation? Should I just use

```
  annotations:
    quarks.cloudfoundry.org/json-value: "true"
```

Although it feels a bit weird that the values would still be treated as JSON even if the annotation is set to `"false"`.

PR #1288, opened 2021-02-19
Remove startup-ordinal, use pre-generated job instances (affects bootstrap, etc)

manno:
[#176906742](https://www.pivotaltracker.com/story/show/176906742)


manno:
Merged despite failing test, this version can't upgrade deployments from quarks-operator 7.1.3 because that has the statefulset-name label. (statefulset's label selector can't be updated after creation).
By avoiding that label, we ensure updates from 6.x work.

Issue #1311, opened 2021-05-17
upgrade of kubeCF from version 2.7.12 to 2.7.13 fails when using "helm upgrade"


deepaksharma009:
**Describe the bug**
"cf-operator-pre-upgrade-hook" is failing (during helm upgrade of kubeCF) with image pull back error while pulling one image from the private image registry of ours though I have provided the correct credentials in values.yaml already for private regsitry access.

**To Reproduce**
Install kubeCF-v2.7.12 and then run helm upgrade command to upgrade to kubeCF-v2.7.13, during upgrade the image pull down from the private registry fails. The error happens during pulling down the kubcf-kubectl image says that invalid credentials for accessing the private registry in order to download the image. Though the same credentials and everythig is same during the install of kubeCF-v2.7.13 independently (during "helm install" of kuneCF-v2.7.13). 

**Expected behavior**
I should be able to upgrade the kubeCF from version 2.7.12 to 2.7.13 using helm upgrade command

**Environment**
 - CF-Operator Version: 7.2.1+0.gaeb6ef3
 - kubeCF version 2.7.13
 
**Additional context**
I have done installation of kubecf version 2.7.12 as well as 2.7.13 separately and successfully and it did worked fine in both cases. but simply upgrade from kubeCF-v2.7.12 to kubeCF-v2.7.13 fails (using "helm upgrade".

jandubois:
Yes, this is a bug. The pull-secret is only defined for the operator service account, but not for the hook service accounts.

You need to copy the logic from `templates/service-account-pull-secret.yaml` into the `templates/service-account-hooks.yaml` (and adjust the names accordingly, of course).

If you have the skills, please file a pull request after testing the change!

PR #1325, opened 2021-06-15
Fix failing helm upgrade 

jbuns:
## Motivation and Context

Fix failing helm upgrade when using `.Values.global.image.credentials`. 

Fixes [#1311](https://github.com/cloudfoundry-incubator/quarks-operator/issues/1311)

## Checklist:

Check item if necessary.

- [ ] Review PRs of changed Quarks dependencies

### DIFF
--- a/deploy/helm/quarks/templates/service-account-hooks.yml
+++ b/deploy/helm/quarks/templates/service-account-hooks.yml
@@ -13,4 +13,17 @@ metadata:
     "helm.sh/hook": {{$hook}}
     "helm.sh/hook-weight": "-2"
     "helm.sh/hook-delete-policy": before-hook-creation
+{{- if $.Values.global.image.credentials }}
+imagePullSecrets:
+- name: {{$hook}}-pull-secret
+---
+apiVersion: v1
+kind: Secret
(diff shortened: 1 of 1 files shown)


deploy/helm/quarks/templates/service-account-hooks.yml
```diff
@@ -13,4 +13,17 @@ metadata:
     "helm.sh/hook": {{$hook}}
     "helm.sh/hook-weight": "-2"
     "helm.sh/hook-delete-policy": before-hook-creation
+{{- if $.Values.global.image.credentials }}
+imagePullSecrets:
+- name: {{$hook}}-pull-secret
```
jandubois, line 18: I don't think you really need to define additional secrets, you can just reference the operator service account secret:

```
- name: {{ template "cf-operator.serviceAccountName" . }}-pull-secret
```

I don't care too much, but generally dislike redundant config. Just makes debugging harder than it needs to be. :)

manno:
@jbuns I released https://github.com/cloudfoundry-incubator/quarks-operator/releases/tag/v7.2.7 with the fix

jbuns:
thanks, @manno!

